PR #24, opened 2026-04-29
--allow-cache-exec

JulianNymark:
feat: add --allow-cache-exec and --allow-cache-exec-any flags

my setup is [opencode](https://github.com/anomalyco/opencode/) + [oh-my-openagent](https://github.com/code-yeongyu/oh-my-openagent) + MCP for [open-websearch](https://github.com/Aas-ee/open-webSearch).

I think it's good to be able to opt out of pieces of this utility, so you can still get the benefits of other parts (like the `.env` and out of scope access sandboxing, though I want to open up some cache execution) instead of it being all or nothing. This seems to be the thinking here already, but it is missing a way to allow cache exec on a per directory basis (or entirely with `--allow-cache-exec-any`, though not recommended)

I was able to get this working on my machine using the following config:
```toml
[allow]
read = ["~/Repos/configs/opencode", "~/.cache/opencode", "~/.config/git", "~/.bun"]
write = ["~/.local/state/opencode", "~/.cache/opencode", "~/.config/opencode", "~/.claude"]
[sandbox]
allow_cache_exec = ["ms-playwright", "opencode"]
```

sandboxing works great! 🎉 (screenshot shows MCPs & orchestration setup working + sandboxing directories)
image


Starefossen:
Took a pass at this — rebased on main and added a few things:

- `../Applications` in `allow_cache_exec` passes validation but SBPL normalises `../` so it escapes `~/Library/Caches`. Now rejecting `..` and `.` components (slash stays allowed for `pnpm/dlx` etc.)
- Extended the char blocklist to match what the rest of the codebase uses (`\r`, `\0`, `(`, `)`, `;`)
- `validate_config_paths()` in `sandbox.rs` didn't validate these strings — added the same check there as a second gate
- The comment saying "No write access" in `sandbox_profile.rs` was wrong — `~/Library/Caches` already has `write: true` in `HOME_TOOL_DIRS`
- `cplt config check` now warns on `allow_cache_exec_any = true` like the other dangerous flags
- Added tests for traversal rejection, unsafe chars, rule ordering, and the diagnostic warning

Also resolved the conflicts with main (`allow_docker` landed in the same spots).

PR #55, opened 2026-05-13
fix: emit sensitive project file denies after all user allows

JulianNymark:
### Bug
When a user configured `allow.read = ["~/Repos"]` in `~/.config/cplt/config.toml`, `.env` files inside projects under `~/Repos` became readable.
**Root cause:** SBPL (macOS Seatbelt) uses last-match-wins semantics. The `.env`/`.pem`/`.key` deny rules were emitted early in the profile (inside `emit_project_access`), while user-configured `allow.read` paths were emitted later by `emit_user_allows`. The later allow overrode the earlier deny.
### Fix
Extract sensitive project file denies into a new `emit_sensitive_project_denies()` function and call it at the **very end** of `generate_profile()`, after all allows (including user-configured paths).
### New rule ordering
... all allows first (project, home, tools, user-configured) ...
... other denies (git hooks, dotfiles, etc.) ...
... network rules ...
(deny file-read*  (regex #"/\.env$"))      ← LAST, so it wins
(deny file-write* (regex #"/\.env$"))
(deny file-read*  (regex #"/\.env\..*"))
...
### Test
Added `profile_env_deny_comes_after_user_allows` which creates a profile with `extra_read: ["/projects"]` and asserts that `.env` deny rules appear **after** the user allow in the generated SBPL string.
### Files changed
- `src/sandbox_profile.rs` — extract denies, move call to end
- `tests/unit_tests.rs` — regression test

### DIFF
--- a/src/sandbox_profile.rs
+++ b/src/sandbox_profile.rs
@@ -116,7 +116,7 @@ pub fn generate_profile(opts: &ProfileOptions) -> String {
 
     emit_header(&mut sb, &project);
     emit_process_rules(&mut sb);
-    emit_project_access(&mut sb, &project, opts.allow_env_files);
+    emit_project_access(&mut sb, &project);
     emit_home_access(&mut sb, &home, opts.agent, opts.agent_dirs);
     emit_git_hooks(&mut sb, opts.git_hooks_path);
     emit_system_access(&mut sb, &home, opts.allow_browser, allow_chromium_runtime);
@@ -153,6 +153,10 @@ pub fn generate_profile(opts: &ProfileOptions) -> String {
         opts.proxy_port,
         opts.localhost_ports,
     );
+    // Sensitive project file denies MUST come after all user-configured allows.
+    // SBPL uses last-match-wins, so a user allow like `allow.read = ["~/Repos"]`
+    // would override the .env deny if emitted before it.
+    emit_sensitive_project_denies(&mut sb, opts.allow_env_files);
 
     sb
 }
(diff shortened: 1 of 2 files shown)


Starefossen (approved):
## Review: ✅ Approve (with one follow-up suggestion)

**The fix is correct.** SBPL last-match-wins semantics mean moving sensitive file denies to the very end of `generate_profile()` is the right approach. The test proves the ordering property directly.

### Verified
- Clean rebase onto `main` (no conflicts)
- All 518 tests pass (188 unit + 330 lib), clippy clean
- The new `profile_env_deny_comes_after_user_allows` test covers the exact scenario from the bug report

### One follow-up (non-blocking, same class of bug)

The `.cplt.toml` write-deny and `.git/hooks` write-deny are still in `emit_project_access` (emitted early, line 221-230). These have the **same vulnerability**: if a user has `allow.write = ["~/Repos"]`, SBPL last-match-wins would let the agent overwrite `.cplt.toml` or write git hooks.

**Config tampering risk:** Agent writes `allow_env_files = true` to `.cplt.toml` → next session the sandbox is weakened.

**Suggested follow-up:** Move `.cplt.toml`, `.git/hooks`, `.git/config`, and `.gitmodules` write-denies into `emit_sensitive_project_denies()` (or a sibling function) so they also come last. This can be a separate PR.

```rust
fn emit_sensitive_project_denies(sb: &mut String, project: &str, allow_env_files: bool) {
    // Config tamper protection
    sbpl!(sb, "(deny file-write* (literal \"{project}/.cplt.toml\"))");
    sbpl!(sb, "(deny file-write* (subpath \"{project}/.git/hooks\"))");
    sbpl!(sb, "(deny file-write* (literal \"{project}/.git/config\"))");
    sbpl!(sb, "(deny file-write* (literal \"{project}/.gitmodules\"))");

    // .env / .pem / .key denies...
    if !allow_env_files { ... }
}
```

### Summary
Good catch on the last-match-wins semantics bug. The fix is minimal, well-documented, and properly tested. 🎉

Starefossen (approved):

PR #66, opened 2026-05-15
fix: handle SEA extraction for npm-installed copilot

### LINKED ISSUE #27: OAuth code flow with browser
Opening a browser seems to be blocked? For eg. and MCP server where OAuth code flow is used, one can start `copilot` outside sandbox, do auth, and later started `cplt` will use the tokens, but it needs to be done every time the tokens have expired. 

Is the security concern in `copilot` triggering a code flow and retrieving a token it should not have by opening default browser where the user is logged in?

Starefossen:
## Problem

When Copilot is installed via npm (`npm i -g @github/copilot`), the binary in PATH is a **node wrapper script** — not a Mach-O SEA binary. The previous `is_macho_binary()` check returned `false` and skipped pre-sandbox extraction entirely.

But npm-installed Copilot still does SEA extraction at runtime: the `npm-loader.js` bootstraps the embedded SEA which extracts to `~/Library/Caches/copilot/pkg/darwin-arm64/`. Since the sandbox denies writes to this path (write-then-exec defense), users see:

```
Failed to extract bundled package: Error: EPERM: operation not permitted, mkdir '.../.extracting-1.0.41-...'
```

## Fix

**Remove the Mach-O vs non-Mach-O distinction entirely.** The `binary_identity()` function (canonicalized path + inode + size + mtime) works for any file type — Mach-O binaries, shell scripts, symlinks. All copilot installs now use the same fast-path cache:

1. Compute identity from file metadata of the resolved copilot binary
2. If identity matches cache AND extraction dir still exists → skip (fast path)
3. Otherwise → trigger extraction outside sandbox → persist identity + dir name

**This also fixes the version upgrade edge case**: when `npm update` reinstalls copilot, the wrapper script's mtime changes → cache invalidates → extraction triggers for the new version.

## Behavior matrix

| Install method | Before | After |
|---|---|---|
| Homebrew Cask (Mach-O) | ✅ Fast-path cache | ✅ Same behavior |
| npm global (node script) | ❌ EPERM | ✅ Fast-path cache |
| npm upgrade (version change) | ❌ EPERM (stale cache) | ✅ Cache invalidates, re-extracts |
| Binary deleted from disk | Stale cache, re-extracts | Same — marker check fails, re-extracts |

## Testing

- 330 unit/lib tests pass
- 41 integration tests pass  
- 110 e2e tests pass (1 pre-existing flaky trust test unrelated)
- 39 e2e_projects tests pass

Closes #27

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,4 @@ src/test/
 tests/test_new.py
 .DS_Store
 .idea
+.cplt-fake-trust-locked-*/
--- a/src/main.rs
+++ b/src/main.rs
@@ -2891,28 +2891,29 @@ fn ensure_copilot_extracted(copilot_bin: &Path, home: &Path) -> Result<(), Strin
         _ => return Ok(()),
     };
 
-    // Skip extraction for non-Mach-O binaries (e.g. shell script wrappers).
-    // SEA extraction only applies to the compiled Copilot Mach-O binary.
-    if !is_macho_binary(copilot_bin) {
-        return Ok(());
-    }
-
-    let Some(binary_id) = binary_identity(copilot_bin) else {
-        return Ok(());
-    };
-
     let pkg_base = home
         .join("Library/Caches/copilot/pkg")
         .join(format!("darwin-{arch}"));
+
+    // Compute binary identity for the fast-path cache (only works for Mach-O).
+    // npm-installed copilot uses a node/shell wrapper — not Mach-O, so binary_id
+    // will be None. We still need to handle extraction for npm installs.
+    let binary_id = if is_macho_binary(copilot_bin) {
+        binary_identity(copilot_bin)
+    } else {
+        None
+    };
 
     // Fast path: check cplt-managed marker that records both the binary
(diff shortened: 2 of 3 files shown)


Starefossen:
Both review issues are addressed in the latest commits:

1. **Version upgrade edge case** (stale `find_any_complete_dir`): Fixed by removing the non-Mach-O fast-path entirely. Now all installs use `binary_identity()` (path + inode + size + mtime) as cache key. When npm upgrades copilot, mtime changes → cache invalidates → extraction triggers correctly.

2. **Editor shim concern**: Fixed by gracefully handling non-SEA binaries. If `copilot --version` exits cleanly but produces no extraction and `pkg_base` has no version dirs, we return Ok (nothing to extract). This handles editor shims, dev builds, and test fakes.

Net result: simpler code (-34 lines from first commit), unified identity for all copilot install types, and robust handling of edge cases.

Starefossen:
@copilot ci is still failing, please debug and fix it

PR #92, opened 2026-06-10
fix(proxy): allow --allow-localhost ports to bypass proxy on Linux

nilsmsa:
## Problem

On Linux, `curl http://localhost:5173/` inside a cplt session launched with `--allow-localhost 5173` returned `405 Method Not Allowed`. Outside cplt it worked fine.

**Root cause:** Two separate layers both blocked the connection:

1. **Plain HTTP 405**: `curl` routes plain HTTP through the cplt CONNECT proxy (via `HTTP_PROXY`). The proxy only handles HTTPS tunneling (CONNECT method) and returns 405 for plain GET.
2. **`NO_PROXY` not set on Linux**: On macOS, `NO_PROXY=localhost,...` is always set because Seatbelt enforces localhost isolation at the kernel level. On Linux, it was intentionally omitted — the proxy was supposed to be the sole localhost enforcer. But when the user explicitly opens a port with `--allow-localhost`, the proxy should step aside.

## Fix

**`sandbox_exec.rs`**: Set `NO_PROXY=localhost,127.0.0.1,::1` on Linux when the user has configured `--allow-localhost` or `--allow-localhost-any`. This lets tools like `curl` connect directly, bypassing the proxy entirely. Without explicit localhost access configured, the proxy continues to block all loopback connections on Linux.

**`proxy.rs` + `main.rs`**: Defense-in-depth carve-out in `handle_connect`: when the target is localhost and the port is in `allow_localhost_ports` (or `allow_localhost_any` is set), skip the private-IP block. This ensures HTTPS CONNECT tunnels to localhost also work for tools that ignore `NO_PROXY`.

## Tests

- Unit test: verifies the carve-out logic for allowed vs. blocked localhost ports
- Integration (Linux): `allow_localhost_port_permits_tcp_connect` — confirms TCP connect succeeds on the allowed port
- Integration (Linux): `allow_localhost_port_does_not_open_other_ports` — confirms non-allowed ports remain blocked

## Verified

```
curl -v http://localhost:5173/
# HTTP/1.0 200 OK  ✓
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -702,6 +702,17 @@ cplt config explain   # list all keys with descriptions
 
 Per-repo config (`.cplt.toml`) operates as a separate layer: `[deny]` tightens unconditionally, and approved permissions are **additive only** — they can enable features but cannot disable anything set by CLI or global config.
 
+#### Global-only settings
+
+These keys stay in `~/.config/cplt/config.toml` and are rejected from
+`.cplt.toml` because they are machine-specific or local preferences:
+
+`sandbox.agent`, `sandbox.quiet`, `sandbox.yes`, `sandbox.validate`,
+`sandbox.scratch_dir`, `sandbox.pass_env`, `sandbox.inherit_env`,
+`sandbox.allow_cache_exec`, `sandbox.allow_cache_exec_any`, `proxy.enabled`,
+`proxy.port`, `proxy.log_file`, `proxy.log_level`, `proxy.blocked_domains`,
+`proxy.allowed_domains`, and all `[gh_guard]`, `[git_guard]`, and `[audit]` keys
+
 ### Per-repo configuration (`.cplt.toml`)
 
 Commit a `.cplt.toml` to your repository root to enforce team policy:
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1,5 +1,16 @@
 # Configuration
 
+## How cplt config works
+
+cplt reads config in this order:
+
+1. CLI flags for the current run
+2. `~/.config/cplt/config.toml` (or `CPLT_CONFIG`)
+3. built-in defaults
+
+List values are merged, so repeated `cplt config set` commands accumulate for
+`allow.read`, `allow.write`, `allow.ports`, `allow.localhost`, and `deny.paths`.
+
 ## Quick setup
 
 Use `cplt config set` to configure cplt without editing files:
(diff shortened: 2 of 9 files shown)


nilsmsa:
Ikke helt god denne her, ser mer på det senere.

Starefossen (changes requested):
## 🔴 HIGH: DNS rebinding via `*.localhost` bypasses the post-DNS private-IP guard

**Location:** `src/proxy.rs` line 614

`localhost_connect_allowed` is computed from the hostname *before* DNS resolution, using `h.ends_with(".localhost")`. This flag then suppresses the post-DNS check:

```rust
if is_private_ip(&socket_addr.ip())
```

The comment at line 610 explicitly says this block *"prevents DNS rebinding attacks"* — but the bypass disables it.

**Attack path:**
1. Agent issues `CONNECT evil.localhost:PORT` through the proxy
2. `evil.localhost` resolves to `169.254.169.254` (cloud IMDS) or `10.0.0.1` (internal)
3. `h.ends_with(".localhost")` → `localhost_connect_allowed = true`
4. Pre-DNS hostname check: bypassed. Post-DNS IP check: bypassed.
5. Proxy tunnels to cloud metadata → IAM credential exfiltration via `is_link_local()` range

**Fix — verify the resolved address is actually loopback:**

```rust
if is_private_ip(&socket_addr.ip())
{
    log_connection("CONNECT", target, "BLOCKED-PRIVATE-RESOLVED", ...);
}
```

This preserves the intended behaviour (`localhost:5173` resolves to `127.0.0.1` → loopback confirmed → allowed) while keeping the rebinding guard if `evil.localhost` resolves to any non-loopback private IP.

---

## 🟡 MEDIUM: `NO_PROXY` bypasses proxy for all loopback when any single port is opened

**Location:** `src/sandbox_exec.rs` lines ~127–131

When `--allow-localhost 5173` is set, `NO_PROXY=localhost,127.0.0.1,::1` is injected for the entire session — not just port 5173. This removes the proxy's audit/block layer for all loopback. On Linux, Landlock enforces per-port isolation at the kernel level, so unapproved ports are still blocked. But if Landlock is absent (kernel < 5.13) or if proxy logs are relied on for audit, the bypass is silent and broader than intended.

---

## ✅ Non-issues

- `parse_repo_from_url` with embedded CI token: only `owner/repo` path extracted, token not logged ✅
- IPv4-mapped loopback (`::ffff:127.0.0.1`): handled by `is_v4_mapped_private` ✅
- Direct numeric loopback IPs: `h.parse::<IpAddr>().is_ok_and(|ip| ip.is_loopback())` correctly restricts to actual loopback addresses ✅
- Domain allowlist not bypassed by `localhost_connect_allowed` ✅
- Test coverage: 4 real proxy CONNECT tests + 2 Linux integration tests ✅

Starefossen (approved):

PR #104, opened 2026-06-26
fix(init): use = instead of += in config hints

nilsmsa:
## Problem

`cplt init` suggested adding lines like:

```toml
allow.read += ["~/.gradle/gradle.properties"]
```

to the personal config (`~/.config/cplt/config.toml`). The `+=` operator is not valid TOML syntax, so `toml::from_str` rejects it and cplt fails to start.

## Fix

Replace `+=` with `=` in all suggestion display strings and personal config hint comments. Three locations in `src/init.rs`:

- `format_suggestion()` — terminal display in `cplt init` report
- `personal_hints` in `generate_toml()` — comments in generated `.cplt.toml`
- `format_global_suggestion()` — terminal display in `cplt init --global`

## Why `=` is equivalent

`allow.read` in the personal config is purely additive — it extends cplt's internal sandbox rules via `.extend()`, never replacing them. Repo config (`propose.allow.read`) is also merged additively on top. Using `=` with a single array entry is fully correct.

### DIFF
--- a/src/init.rs
+++ b/src/init.rs
@@ -135,17 +135,17 @@ fn write_detection(out: &mut String, d: &Detection) {
 fn format_suggestion(s: &Suggestion) -> String {
     match s {
         Suggestion::Propose(flag) => format!("{} = true", flag.key_name()),
-        Suggestion::AllowRead(p) => format!("allow.read += [\"{p}\"]"),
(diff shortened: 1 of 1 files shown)


Starefossen (approved):

Starefossen:
Thanks 🙏 💯

Issue #114, opened 2026-07-04
Design: kernel-level network isolation for the Linux sandbox (netns + proxy bridge)


Starefossen:
> Design PRD for kernel-level Linux network isolation (netns + proxy bridge). Kept as an issue rather than a docs PR.

# Design: kernel-level network isolation for the Linux sandbox (netns + proxy bridge)

> Status: **proposed** (design only — no code yet). Tracks
> [issue #114](https://github.com/navikt/cplt/issues/114).
> Scope: Linux only. macOS Seatbelt already enforces the equivalent (see below).
>
> Read [SECURITY.md](../../SECURITY.md) (network section) and
> [docs/proxy.md](../proxy.md) (proxy-forced mode) first — this document extends
> both. It is a PRD: it fixes the *what* and the *why*, sketches the *how*, and
> records the decisions. Implementation lands in follow-up PRs.

---

## 1. Problem & goal

On Linux, cplt's outbound-network story is two independent layers:

1. **Landlock TCP-connect rules** (`sandbox_landlock.rs`, ABI v4+ / kernel 6.7+).
   These are **port-based**: a rule is `NetPort::new(port, AccessNet::ConnectTcp)`
   — it allows connecting to *that port on any host*. There is no way to say
   "port 443 only to `127.0.0.1`". See the `NetRule` type
   (`sandbox_landlock.rs`) and `generate_policy()`'s network block.
2. **The cplt CONNECT proxy** (`proxy.rs`), which does the real work: hostname
   allow/block lists, resolved-IP SSRF guard, port policy, upstream chaining.
   The agent is pointed at it via `HTTP_PROXY` / `HTTPS_PROXY` set in
   `sandbox_exec.rs` (`http://127.0.0.1:<proxy_port>`).

The proxy runs on **host loopback** (`proxy::start()` binds `127.0.0.1:<port>`).
For the agent to reach it, Bubblewrap deliberately **does not** create a network
namespace: `build_bwrap_args()` in `sandbox_bubblewrap.rs` emits `--unshare-pid`,
`--unshare-ipc`, `--unshare-uts`, `--unshare-cgroup`, `--unshare-user` but
**never `--unshare-net`** (asserted by the `args_isolate_expected_namespaces`
test). The agent therefore shares the host network namespace — it can see host
loopback, the LAN, and the internet, constrained only by Landlock's port rules.

**Goal.** Provide, on Linux, a **kernel-enforced default-deny egress** where the
[...]

PR #133, opened 2026-07-05
docs(design): Linux netns network isolation PRD (#114)

Starefossen:
Decision-grade design PRD for #114 (kernel-level network isolation on Linux). Design only — no src/ changes; adds `docs/design/linux-netns.md`.

**Recommended approach:** add `--unshare-net` to bwrap and expose the host CONNECT proxy inside the private netns via a **unix-socket bridge** driven by the existing re-entry helper — no new dependency, stays rootless, reuses the audited confirm-pipe machinery. A netns with only `lo` has no route to external hosts, closing the port-based Landlock residual (evil.com:proxy_port) address-completely and reaching macOS SBPL parity; host-loopback services become unreachable.

Covers: why netns (Landlock is port-based), options evaluated (unix-socket bridge vs pasta/slirp4netns vs veth+nftables), DNS (proxy-side resolution), fail-closed fallback to port-based Landlock, a testing plan reusing the CI :443 differential + require-mode, honest security posture, and the SECURITY.md edits to make when implemented. Primary open question flagged: bringing `lo` up unprivileged inside the userns — a Phase-0 spike, with pasta as the escape hatch.

### DIFF
--- a/docs/design/linux-netns.md
+++ b/docs/design/linux-netns.md
@@ -1,0 +1,511 @@
+# Design: kernel-level network isolation for the Linux sandbox (netns + proxy bridge)
+
+> Status: **proposed** (design only — no code yet). Tracks
+> [issue #114](https://github.com/navikt/cplt/issues/114).
+> Scope: Linux only. macOS Seatbelt already enforces the equivalent (see below).
+>
+> Read [SECURITY.md](../../SECURITY.md) (network section) and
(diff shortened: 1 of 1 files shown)


Starefossen:
Design content moved into the tracking issue #114 — keeping design PRDs as issues rather than docs PRs.

Issue #52, opened 2026-05-13
security: default domain allowlist per agent (fail-closed networking)


Starefossen:
## Problem

Port 443 is open at the kernel level (SBPL cannot filter by domain). The proxy supports `--allowed-domains` but it is **opt-in** — by default, any HTTPS endpoint on the internet is reachable. A malicious dependency can exfiltrate source code to `evil.com:443`.

## Proposed solution

Ship a built-in domain allowlist per agent type. When no explicit `allowed_domains` is configured, use the agent-specific default. This makes the proxy fail-closed for unknown domains.

### Copilot agent needs

```
# GitHub Copilot infrastructure
*.githubcopilot.com
api.github.com
github.com
copilot-proxy.githubusercontent.com
*.actions.githubusercontent.com
default.exp2.cds.s9ch.io

# Package registries (detected by cplt init)
registry.npmjs.org
registry.yarnpkg.com
repo.maven.apache.org
plugins.gradle.org
crates.io
static.crates.io
pypi.org
files.pythonhosted.org
```

### Implementation

1. Add `DEFAULT_ALLOWED_DOMAINS` per agent in `src/agent.rs`
2. In proxy startup: if no `allowed_domains` file/config, use agent defaults
3. Add `--allow-all-domains` flag to disable (escape hatch for debugging)
4. `cplt init` detects project dependencies and adds relevant registries to config
5. Proxy logs blocked domain attempts (at `--proxy-log blocked` level) so users see what to add
6. `cplt doctor` shows active domain policy

### UX flow

```
$ cplt
[cplt] Domain policy: 12 domains allowed (copilot defaults + registry.npmjs.org)
[cplt] Use --allow-all-domains or add domains to allowed-domains.txt to permit others
```

When a domain is blocked:
```
[cplt] ⚠ Blocked connection to evil.com:443 (not in allowed domains)
```

## Security gain

Blocks exfiltration to arbitrary HTTPS endpoints. Combined with proxy-forced mode (#proxy-forced), this makes domain filtering unbypassable.

## Related

- #10 — auto-updatable blocklist (complementary: blocklist blocks known-bad, allowlist permits only known-good)
- Proxy-forced mode (future) — makes this kernel-enforced rather than advisory
- #4 — token scoping (reduces what can be stolen even if exfil channel exists)

Starefossen:
Folding into **#334**, which #334 itself proposes and which carries the current analysis.

[...]

PR #140, opened 2026-07-07
feat(proxy): per-agent default allowlist domains for non-Copilot agents (#52)

Starefossen:
Follow-up to #52/#139: gives the non-Copilot supported agents their own built-in infrastructure domains, so opt-in fail-closed networking (`proxy.default_allowlist`) works for them too (previously only Copilot had infra defaults; others got the registry base only and would have their model/auth traffic blocked).

Per-agent (bare domains, exact-or-subdomain matched):
- **Gemini** → Google AI base (generativelanguage / cloudcode-pa / oauth2 googleapis + accounts.google.com) — high confidence.
- **Claude Code** → api.anthropic.com, console.anthropic.com, claude.ai, statsig.anthropic.com — high confidence.
- **Antigravity** → Google AI base + antigravity.google (it's Google's; stores config under ~/.gemini).
- **OpenCode** → opencode.ai, models.dev — its own infra only; **provider-agnostic**, so the user's chosen LLM provider (Anthropic/OpenAI/Google) must be added via `allowed_domains` (documented).
- **Pi** → no infra listed (endpoints not known — honest gap, documented; needs user-added domains).
- **Shell** → registry base only (not an AI agent).

All agents still get the shared package-registry base. **Best-effort + safe:** this is opt-in and every blocked domain is logged (BLOCKED-ALLOWLIST), so a missing endpoint is a one-line add, not a silent failure. Confidence levels are marked in code comments; Pi/OpenCode gaps are explicit. Tests assert each agent's infra + registry base, non-empty, and bare (no glob). clippy clean both targets.

### DIFF
--- a/docs/proxy.md
+++ b/docs/proxy.md
@@ -146,7 +146,7 @@ cplt config set proxy.allowed_domains "~/.config/cplt/allowed-domains.txt"
 
 By default the proxy is **allow-all**: it logs and blocks known-bad domains, but any HTTPS endpoint on the internet is reachable. Because port 443 is open at the kernel level, a compromised agent or a malicious dependency could exfiltrate source code to an arbitrary host. Network control is the only defense-in-depth layer that stops exfiltration *after* an agent is compromised (issue #52).
 
-Each agent ships with a built-in **default allowlist** — the set of domains it legitimately needs. For **Copilot** this is the GitHub Copilot infrastructure plus common package registries:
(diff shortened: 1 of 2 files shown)


Starefossen:
Follow-up (post-merge): the non-Copilot domain lists here are best-effort and offline-unverifiable. Verify/complete them empirically with `cplt --agent <X> --observe-domains -- <representative task>` (from #143) and update src/agent.rs — especially **Pi** (no defaults) and **OpenCode** (provider-agnostic). This is the human-only prerequisite for the secure-by-default Stage 3 flip (#147).

PR #158, opened 2026-07-31
Add basic support for MSBuild in cplt

markus-photoncycle:
Add `--allow-msbuild` for .NET/MSBuild build server sockets

`dotnet build` (as well as other dotnet commands such as `dotnet restore`) runs MSBuild in server mode, which keeps a long-lived compiler process alive and talks to it over a Unix domain socket at `/tmp/MSBuild<pid>`. The sandbox blocks that today, so this adds an opt-in `--allow-msbuild` flag (and a matching `allow_msbuild` config key) that follows the exact same pattern as `--allow-jvm-attach`.

It only allows sockets matching `/tmp/MSBuild<pid>`, everything else in `/tmp` stays blocked.

This was tested only on Mac.


Starefossen:
Thanks for your first contribution—adding .NET/MSBuild support is valuable!

Before merging, we should clarify the IPC being enabled:

- `/tmp/MSBuild<PID>` is used by MSBuild worker nodes ([source](https://github.com/dotnet/msbuild/blob/main/src/Shared/NamedPipeUtil.cs)).
- The persistent MSBuild server uses `MSBuildServer-{hash}` ([documentation](https://github.com/dotnet/msbuild/blob/main/documentation/MSBuild-Server.md#pipe-name-convention--handshake), [source](https://github.com/dotnet/msbuild/blob/main/src/Build/BackEnd/Node/OutOfProcServerNode.cs)).

The current implementation therefore enables worker-node IPC, not the persistent server described in the PR. Allowing the persistent server socket could also let sandboxed code reuse a server started outside the sandbox.

Suggested direction:

1. Disable persistent server reuse with `DOTNET_CLI_DO_NOT_USE_MSBUILD_SERVER=1`.
2. Describe `--allow-msbuild` as enabling worker-node IPC.
3. Add a real `dotnet build` integration test.
4. Update `SECURITY.md` and user documentation with the new socket exception.

The narrow, opt-in socket rule is a good starting point; it mainly needs corrected semantics and coverage.

markus-photoncycle:
@Starefossen, I addressed your comments in the latest commit (I hope that I understood your comments correctly). 

This represents sort of a minimal state that can be used to build upon, there are many features in the dotnet ecosystem that will not work out of the box here (Aspire, more elaborate build setup, etc.). 

But the minimal case `dotnet restore`, `dotnet build` and `dotnet test` works (at least on my project).

Starefossen (dismissed):

Starefossen:
@markus-photoncycle Heads-up: the latest macOS integration job is red in `project_dotnet_build_works_with_allow_msbuild`. The test receives:

```text
RESULT:dotnet_build:FAIL:sandbox_deny
RESULT:dotnet_run:FAIL:build_failed
```

The profile currently permits only `^/private/tmp/MSBuild[0-9]+$`. This may mean the runner creates the worker pipe under its resolved macOS `TMPDIR` (commonly `/private/var/folders/...`) rather than `/private/tmp`, or that a different operation is denied. Could you surface the captured `DOTNET_ERROR`/actual pipe path in the failing assertion and then keep any policy adjustment scoped to the exact observed path? [Failed job](https://github.com/navikt/cplt/actions/runs/30813935175/job/91708810037).

A few non-blocking cleanups worth including while touching the tests:

- `dotnet_available()` only checks that `dotnet --version` succeeds, while the fixture targets `net8.0`; require SDK major >= 8 (and make the skip message distinguish missing vs too old).
- Do not emit `RESULT:dotnet_build:OK:build_error_but_not_sandbox` for an unrelated build failure. Preserve it as `FAIL` with diagnostics so the test cannot present a failed build as success.
- Narrow `!p.contains("MSBuild")` to the exact socket rule, and consistently call `MSBuild<PID>` worker-node IPC rather than the persistent “MSBuild build server”. The config template/registry/CLI help still contain the old persistent-server wording too.

markus-photoncycle:
@Starefossen what can I do to unblock this PR?

Starefossen:
Adversarial review — one security finding, fixed and pushed to this branch.

**Write-then-exec via `allow.write` over `DOTNET_ROOT`.** `emit_dotnet_root` is the only emitter that re-grants `process-exec` inside a tree that is otherwise writable (`$DOTNET_ROOT/{dotnet,sdk,shared}`), and it paired that with `(deny file-write* ...)` right alongside. But SBPL is last-match-wins and `emit_user_allows` runs five emitters later, so a user `allow.write` covering DOTNET_ROOT — `allow.write = ["~/.dotnet"]`, or any parent of it — overrode the write-deny while the exec allow stayed in force. Net effect: the agent can drop a binary into `$DOTNET_ROOT/sdk` and execute it, which is exactly what `allow_tmp_exec` and the `emit_tool_dirs` exec-denies exist to prevent.

Split the write-denies into `emit_dotnet_exec_denies` and moved them after `emit_user_allows`, matching the convention `emit_sensitive_project_denies` already documents in that same function. Added `profile_dotnet_exec_paths_stay_readonly_under_user_allow_write` asserting the ordering — it fails against the previous commit.

The rest holds up well. The `^/private/tmp/MSBuild[0-9]+$` regex is correctly anchored and genuinely can't match `MSBuildServer-<hash>`, and belt-and-braces'ing it with `DOTNET_CLI_DO_NOT_USE_MSBUILD_SERVER=1` is the right call. Note `--allow-msbuild` is a no-op on Linux (Landlock gates the socket path via filesystem rules, not a separate socket capability) — harmless, but might be worth a `ui::warn` like the other Linux-unenforceable options get.

Verified locally: `cargo fmt --check`, `clippy --all-targets -D warnings`, and 266 unit/lib tests all pass.

Starefossen:
Follow-up, not blocking: `--allow-msbuild` is a no-op on Linux — Landlock gates the socket through filesystem rules, so there's no separate socket capability to grant. Harmless, but since this was only tested on Mac, a `ui::warn` like the other Linux-unenforceable options get would avoid implying the flag does something there.

PR #169, opened 2026-08-12
docs: Gradle plugin workers don't inherit JAVA_TOOL_OPTIONS

Starefossen:
Documents the failure mode reported by Geir Harris in sokos-dare-poc: `./gradlew ktlintFormat` fails with `ConnectException: Could not connect to server [... port:55788, addresses:[/127.0.0.1]]` even with `allow_localhost_any = true`.

**Analysis (confirmed against cplt source):**
- cplt injects `-Djava.net.preferIPv4Stack=true` via `JAVA_TOOL_OPTIONS` (a2199a9) to avoid IPv4-mapped `::ffff:127.0.0.1` addresses that SBPL's `localhost:*` filter cannot match
- Gradle plugin workers (ktlint-gradle) fork JVMs with explicit `forkOptions {}` and don't propagate env → the flag is lost
- SBPL has no primitive for IPv4-mapped matching — this cannot be fixed in cplt or the sandbox profile
- The correct fix is upstream: ktlint-gradle must propagate env in `forkOptions` ([JLLeitschuh/ktlint-gradle#1110](https://github.com/JLLeitschuh/ktlint-gradle/issues/1110), PR [#1112](https://github.com/JLLeitschuh/ktlint-gradle/pull/1112))

Adds a `known-impacts.md` section documenting the symptom, root cause, and the `forkOptions { environment(...) }` fix pattern. Docs-only change.

### DIFF
--- a/docs/known-impacts.md
+++ b/docs/known-impacts.md
@@ -336,6 +336,26 @@ If all of that works, `cplt --allow-gpg-signing` will work too. The `gpg-agent` 
 - **`--deny-path` wins.** If you specify `--deny-path ~/.gnupg` alongside `--allow-gpg-signing`, the deny takes precedence — all GPG allows are suppressed.
 - **`GNUPGHOME`** is not supported yet — only the default `~/.gnupg` location is allowed.
 
+## Gradle plugin workers and `JAVA_TOOL_OPTIONS`
+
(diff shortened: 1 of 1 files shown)


Starefossen:
Superseded by #170, which includes this documentation plus the init-script mitigation.

PR #176, opened 2026-08-20
fix: Cache Poisoning via caching of untrusted files 

ybelMekk:
## release.yaml

Removes `actions/cache` from the release build job.

The job is reachable via `workflow_dispatch` on an arbitrary ref, where it runs in that ref's cache scope and restores whatever a prior run on that branch wrote under a `release-*` key. More generally, a workflow that publishes installable, signed, attested binaries should not consume a shared mutable cache, so the fix is to drop the cache rather than to narrow it.

## Trade-off

Release builds go from roughly 2 to 3 minutes warm to 5 to 8 minutes cold. Releases are infrequent, and a reproducible build from a clean `target/` is worth the wait.

## Not changed

`ci.yaml` keeps `target/` in its caches. PR runs write into the `refs/pull/N/merge` cache scope, which no other branch can read, and their keys do not overlap the release job's `release-*` prefix, so there is no path from a PR cache into a privileged build.

## Docs

`SECURITY.md` gains a "Cache poisoning" subsection recording why `release.yaml` runs cacheless and why `ci.yaml` does not need to.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -72,16 +72,6 @@ jobs:
           toolchain: "1.94"
           targets: ${{ matrix.target }}
 
-      - name: Cache cargo registry and build
-        uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # ratchet:actions/cache@v6.1.0
-        with:
-          path: |
-            ~/.cargo/registry
-            ~/.cargo/git
-            target
-          key: release-${{ matrix.target }}-${{ hashFiles('Cargo.lock') }}
-          restore-keys: release-${{ matrix.target }}-
-
       - name: Build release binary
         run: |
           CPLT_LONG_VERSION="${{ needs.version.outputs.version }}" cargo build --release --target ${{ matrix.target }}
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -910,6 +910,23 @@ The GitHub Actions workflow runs in two stages:
 1. **Linux (ubuntu-latest)**: formatting check (`cargo fmt`), linting (`cargo clippy -D warnings`), unit tests
 2. **macOS (macos-latest)**: full test suite including integration tests, release binary build and verification
 
+#### Cache poisoning
+
+The release workflow (`release.yaml`) deliberately uses **no** `actions/cache`
+step. It runs in a privileged context, triggered on the default branch via
+`workflow_run`, with `contents: write`, `id-token: write`, and
+`attestations: write`, and it publishes signed, attested binaries. A cache entry
+populated by a lower-trust run (a `pull_request` build, or a `workflow_dispatch`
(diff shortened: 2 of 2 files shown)


Starefossen:
Reviewed. The `release.yaml` half should land. The `ci.yaml` half is fixing something that cannot happen, and I would drop it.

**Why `ci.yaml` does not need this.** GitHub scopes cache access by ref: a run restores caches created in its own branch or in the default branch, and nothing else. A `pull_request` run writes into the `refs/pull/N/merge` scope, which no other branch can read, so a poisoned entry from a PR run has nowhere to go. It cannot reach `main`, it cannot reach another PR, and it cannot reach the release build. The keys rule it out a second time: the PR jobs write `pr-build-*` and `Linux-integration-*`, the release job restores `release-${{ matrix.target }}-`, and `restore-keys` only matches by prefix. So dropping `target/` from those four steps buys no security and costs a cold compile on every PR and every merge-queue entry.

**Why `release.yaml` should still land.** Not for the reason in the description, but for a better one. `release.yaml` has a `workflow_dispatch` trigger, so anyone with write access can run it on a non-default ref, where it executes in that branch's cache scope and will happily restore a `release-*` key that an earlier run on that same branch planted. Beyond that, a workflow that publishes binaries other people install should not be consuming a shared mutable cache at all, whatever the current reachability analysis says. Trading two minutes of release build time for that is the right trade, and the PR already makes the argument.

Suggested description if you keep only the release change:

> ## release.yaml
>
> Removes `actions/cache` from the release build job.
>
> The job is reachable via `workflow_dispatch` on an arbitrary ref, where it runs in that ref's cache scope and restores whatever a prior run on that branch wrote under a `release-*` key. More generally, a workflow that publishes installable binaries should not consume a shared mutable cache, so the fix is to drop the cache rather than to narrow it.
>
> ## Trade-off
>
> Release builds go from roughly 2 to 3 minutes warm to 5 to 8 minutes cold. Releases are infrequent, and a reproducible build from a clean `target/` is worth the wait.
>
> ## Not changed
>
> `ci.yaml` keeps `target/` in its caches. PR runs write into the `refs/pull/N/merge` cache scope, which no other branch can read, and their keys do not overlap the release job's `release-*` prefix, so there is no path from a PR cache into a privileged build.

Happy to push the revert of the `ci.yaml` hunks if you would rather not redo the branch.

Starefossen:
Closing this as superseded by #274, which removed the same `actions/cache` step from `release.yaml` and merged earlier today.

The duplication was mine, not yours. You filed this on 20 August. I wrote #274 on 3 September without checking for an open PR against the alert first, and it went in ahead of yours. Your diagnosis was the same one and the fix was the same fix, three weeks sooner.

For the record, since the reasoning is worth keeping: dropping the cache rather than narrowing the key is right. A scoped key still writes attacker-influenced bytes into a cache the default branch can restore, so the vector survives the narrowing. Removing it is the only version that closes it.

PR #184, opened 2026-08-24
test(proxy): poll for observed domains instead of sleeping

Starefossen:
The two `observe_*` proxy tests snapshot `observed_domains()` right after `proxy_connect` returns, but the connection thread records the host *after* the client call completes. A fixed `sleep(50ms)` papered over that race.

It overshot on a loaded runner during #158's CI and failed the run on a test unrelated to that PR:

```
thread 'proxy::tests::observe_allow_all_records_domain_that_would_be_blocked'
panicked at src/proxy.rs:2964:14:
contacted host must be recorded even in allow-all mode
test result: FAILED. 703 passed; 1 failed
```

Replaces both sleeps with an `await_observed` helper that polls to a 5s deadline — the same retry idiom `assert_connect_allowed` already uses a few lines above. Faster in the common case (5ms granularity instead of a flat 50ms) and reliable under load.

Verified: `cargo fmt`, `clippy -D warnings`, 173 proxy tests pass, and the two `observe_*` tests run 20/20 clean.

### DIFF
--- a/src/proxy.rs
+++ b/src/proxy.rs
@@ -4338,11 +4338,9 @@ mod tests {
             Some(resolver),
         );
 
-        let status = proxy_connect(proxy.port, "intranet.corp.example:443");
-        assert!(
-            status.contains("200"),
-            "allow_private_domains host must be forwarded to upstream; got: {status}"
-        );
+        // Retry transport flakes: `assert_connect_allowed` still fails fast on a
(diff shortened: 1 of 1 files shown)


Starefossen:
Folded into #175 rather than shipping separately — both commits (`await_observed` polling helper, and routing the three bare CONNECT-expects-200 asserts through `assert_connect_allowed`) are now on that branch.

Closing per @Starefossen.

Issue #216, opened 2026-08-27
Design: what multi-repo support should be — the merge model in #165 does not survive contact with the code


Starefossen:
#165 proposes a `--workspace` flag that reads N repos' `.cplt.toml` files and applies the intersection of their allow-lists and the union of their deny-lists. Having mapped what `project_dir` actually controls, that model cannot be built as written. This issue replaces its "Proposed scope": it separates the use cases, proposes something much smaller, and says what is deliberately left undone.

## Why the merge model fails

**Intersection of allow-lists is empty by construction, not merely strict.** Since #203, a relative path in a repo `.cplt.toml` anchors to that repo's own root (`resolve_repo_path`, `src/config/loading.rs:1040`), and `resolve_repo_allow_path` (`:1206-1219`) drops any relative entry resolving outside its repo. Two repos' relative allow-lists are therefore disjoint sets of absolute paths. Intersecting them yields nothing. The containment rule is a good property and should stay — it just makes intersection meaningless.

**Intersection is the wrong operation for everything that is not a path.** `sandbox.*` proposals are booleans (`src/config/loading.rs:1057-1090`). ANDing them across roots turns off `allow_docker`, `allow_localhost_any`, `allow_jvm_attach` and `gradle_init` for *every* root as soon as one root did not ask — exactly the asymmetric need polyrepo work creates. Frontend wants localhost, backend wants Docker; the intersection gives you neither. Ports are the same shape.

**`[propose]` has mixed polarity, so no single merge direction is correct.** Two of its keys *enable* guards: `gh_guard` and `git_push_prevention` set `enabled = true` (`src/config/loading.rs:1091-1098`). Intersecting the section would disable a guard a repo explicitly asked to switch on. No polarity is simultaneously right for `allow_docker` and for `git_push_prevention`.

Worth noting: **cplt contains no intersection logic anywhere today.** Every config layer is additive — `apply_repo_config` appends and dedupes (`:1101-1156`), booleans only go `false → true`. The merge model would be a new kind of operation, and it is the wrong one.

## Demand: what the tracker actually shows

The issue asserts developers have asked for this. I looked, and could not substantiate it:

- `navikt/cplt` has 95 issues, 30 from people other than the maintainer. None concerns multi-repo, polyrepo, workspaces, or sibling repos.
[...]

Issue #210, opened 2026-08-27
The parent-side audit runs git in the project repo outside the sandbox, so a repository can influence what the launcher executes


Starefossen:
## What happens

The post-session change audit runs `git` **in the parent process, outside the sandbox**, with the working directory set to the project being sandboxed.

`src/main.rs:2519`:

```rust
// Run agent inside sandbox, wrapped in the post-session audit lifecycle.
// The audit is on by default and suppressed by --no-audit or --quiet. The
// baseline is captured just before exec and the report printed just after,
// both in the parent (outside the sandbox) — see audit::run.
let audit_enabled = resolved.audit && !resolved.quiet;
let exit_code = audit::run(&project_dir, audit_enabled, || {
    sandbox::exec_sandboxed(...)
});
```

`src/audit.rs`, `git_output`:

```rust
let mut child = Command::new("git")
    .arg("--no-optional-locks")
    .args(args)
    .current_dir(project_dir)
    ...
```

The invocations are `rev-parse HEAD` and `status --porcelain -z -uall` (baseline, **before** exec) and `status`/`diff --numstat` (report, after exit). Baseline capture therefore runs on *every* sandboxed start, not only when a session ends.

`git` run inside a repository honours that repository's configuration, and several git configuration keys name a program git then executes. `core.fsmonitor` is the direct one — `git status` runs it. `.gitattributes` diff/textconv drivers resolve through `diff.<driver>.textconv`, reached by `git diff`. None of `GIT_CONFIG_NOSYSTEM`, `GIT_ATTR_NOSYSTEM`, `-c core.fsmonitor=false`, or `-c protocol.*` hardening is set on these invocations, and nothing constrains them the way the sandboxed child is constrained.

So a repository can influence what the *parent* — the process that establishes the sandbox — executes, and that execution is unsandboxed.

## Why this is being filed now

nav-pilot ([`navikt/copilot`](https://github.com/navikt/copilot)) launches Copilot and OpenCode through cplt. For its new staged-payload launch path it passes `--no-audit` on **every** invocation, and that flag is load-bearing rather than cosmetic. Team eSyfo, who own the reference launcher nav-pilot is being measured against, stated the reason in [navikt/copilot#437](https://github.com/navikt/copilot/issues/437#issuecomment-5437575432):

[...]

PR #217, opened 2026-08-27
fix(sandbox): protect git-persistence paths in every writable root, and stop the audit claiming clean

### LINKED ISSUE #212: A sibling repo granted via allow.write has unprotected .git/hooks, so a planted hook runs unsandboxed
Under `--allow-write <sibling-repo>`, the sibling repo's git-persistence paths are unprotected, so a sandboxed agent can plant a hook there that runs **unsandboxed** on the user's next git operation in that repo.

## Why

The `.git/hooks`, `.git/config`, `.gitmodules` and `.cplt.toml` write-denies are emitted with a literal `{project}` prefix — `src/sandbox_profile.rs:262-285` takes a single project path. Same for the bwrap read-only re-binds (`src/sandbox_bubblewrap.rs:483-500`). Neither loops over granted paths.

So for `project_dir` = A, cplt correctly denies `A/.git/hooks`. With `--allow-write B`, `B/.git/hooks/post-commit` is writable — and hooks run outside the sandbox by definition.

This is the exact vector cplt documents blocking (`README.md:87`, `docs/security.md`). It is simply absent for any second root.

## Scope

- `--allow-write` / `allow.write` pointing at a git repo
- `[propose.allow] write` approved via `cplt trust accept`
- Starting cplt from a shared parent directory (`cplt --project-dir ~/code`), where **every** child repo's hooks are writable — see the sibling issue on workspace UX

`--allow-read` is not affected (read-only).

## Fix sketch

Emit the git-persistence denies for every writable granted path that looks like a git repo, not only `project_dir`. `discover.rs` already resolves a repo's common `.git` dir (`git_common_dir`, `src/discover.rs:958`) for the worktree case, so the mechanism exists — it just runs for one root.

Worth deciding at the same time whether `--allow-write` at a git repo should require an explicit opt-in, since it silently disables protections the user would assume apply everywhere.

Found while mapping multi-repo constraints for #165.

### LINKED ISSUE #214: The audit reports a clean session when the agent only changed a repo granted via allow.write
When the agent works in a repo granted via `allow.write` rather than in `project_dir`, the post-session audit prints **"no project file changes"** — an affirmative clean bill of health for a session that rewrote another repository.

## Why

`audit::run(&project_dir, …)` (`src/main.rs:2524`, `:3080`) runs all three git commands with `.current_dir(project_dir)` (`src/audit.rs:551`): `rev-parse HEAD` (`:286`), `status --porcelain -z -uall` (`:294`, `:319`), `diff --numstat` (`:385`). Nothing iterates `resolved.allow_write`.

The module has honest failure states — `AuditReport::Unavailable` (`audit.rs:452`) and `Incomplete` (`:463`) — but both fire only on git *errors*. A grant that puts a writable repo outside the audit's view is not an error, so the clean path is taken and `"no project file changes"` is printed (`audit.rs:495`).

## Why it matters more than the wording suggests

The audit exists to make "here is what the session actually changed" credible — it is the feature that makes the sandbox's claims checkable after the fact. Reporting clean for a session that modified another repo is worse than reporting nothing: it converts an absence of information into a false assurance.

This is the same class as the silent non-enforcement bugs fixed in #203 and #210 — the rule is present in the model and absent in reality, and nothing says so.

## Fix sketch

Two viable directions, and the cheap one may be enough:

1. **Audit every writable granted root**, not just `project_dir`. Correct, and it is what a user would expect, but it needs the git-hardening choke point from #210 applied per root and a report format that names which repo changed.
2. **Say what was not audited.** If `allow_write` contains a path outside `project_dir`, degrade to `Incomplete` or append a line naming the unaudited roots. Much smaller, and it removes the false assurance, which is the actual harm.

Doing 2 now and 1 with the workspace work in #165 seems right, but that is a design call.

Found while mapping multi-repo constraints for #165.

Starefossen:
Closes #212. Closes #214.

Two halves of the same problem: a writable second root sits outside cplt's model, so its git-persistence paths are unprotected (the security half) and the audit reports a clean session for changes made there (the honesty half).

Both are reachable **without any workspace feature** — a repo may propose `allow.write = ["~/code/other-repo"]` absolutely, and the containment rule from #203 only constrains *relative* entries.

## #212 — protections per writable root

**One resolver, reused.** `discover::git_dir_of(dir)` is extracted from the existing `git_common_dir`, which now delegates to it. It returns the directory that actually holds `hooks/` and `config`: `<repo>/.git` for a normal repo, the main repo's `.git` for a worktree, the repo dir itself for a bare repo, the enclosing repo's `.git` for a path inside one, and `None` for a non-repo — a no-op, never an error. `git_common_dir` keeps its own unsafe-root and `$HOME` filters because it *grants* access; `git_dir_of` needs none, because its result only places denies.

**One choke point.** `sandbox::prepare()` computes the resolved dirs and hands them to both `prepare_impl` arms, so every `SandboxConfig` caller — both `main.rs` sites, `check`, `--print-profile` — is covered with no new field to populate.

**Two layers.** The `{root}/.git/hooks|config`, `{root}/.gitmodules`, `{root}/.cplt.toml` denies are emitted for the project *and* every `allow.write` grant, unconditionally rather than only for paths that are repos today — so a repo the agent creates mid-session with `git init` **at the grant root itself** is covered. On top, the resolved `.git` dirs get `hooks`/`config` denies, which is the only thing that reaches a granted worktree or bare repo.

An earlier revision of this description claimed mid-session `git init` was covered generally. Adversarial review reproduced the limit: granting `~/work` denies `~/work/.git/hooks` but leaves `~/work/proj/.git/hooks` writable, whether `proj` was a repo at launch or the agent creates one there. Same shape as the contains-repos gap below; the docstring now names it as a known gap.

Verified against a real `--print-profile` with a worktree and a bare repo granted: the worktree grant produced a deny on the *main* repo's `.git/hooks`, the bare grant on `<root>/hooks`.

### What each platform can actually express

| | project | granted root |
|---|---|---|
| macOS Seatbelt | hooks + config + .gitmodules + .cplt.toml | **same, now** — full parity |
| Linux Landlock alone | **cannot** deny a subpath inside a writable root | **cannot**, same reason — unchanged and unchangeable at this layer |
| Linux Bubblewrap | ro-binds `.git/hooks` + `.cplt.toml` (+ resolved common-dir hooks) | **same, now** |

`.git/config` and `.gitmodules` stay writable under bwrap by a pre-existing deliberate decision — read-only binding breaks `git config`/`remote`/`submodule` and risks a stale `config.lock`.

So on Linux without bwrap a granted repo's hooks are unprotected — exactly as the project's already are (`src/sandbox_landlock.rs:708-717` says so in the code). `docs/security.md` and the README now say "every writable root" rather than "the project", and carry this table.

## #214 — the audit stops asserting a clean session

`audit::run` now takes `&resolved.allow_write`. Any grant not under `project_dir` is named after the report:

```
[cplt] Session ended (exit 0, 0s)
[cplt] no project file changes
[cplt] audit covers /…/project only — 1 writable path outside it was NOT audited: /…/sibling
```

Kept outside `AuditReport` deliberately: the fact is true of all three variants and none has a place to carry it, so the enum shape is untouched. Grants *inside* `project_dir` are omitted, since the project's own `git status` covers them.

This is #214's smaller option — remove the false assurance. Auditing every writable root properly belongs with the workspace work (#216) and needs #210's hardened invoker applied per root first.

## Mutation checks — four, all caught

| mutation | result |
|---|---|
| `extra_git_dirs()` returns empty | `granted_bare_repo_hooks_blocked` fails with `bare_hook_write:OK` |
| `writable_roots(project, &[])` — project only | `granted_sibling_repo_git_hooks_blocked` fails with `git_hook_write:OK`, `git_config_write:OK`, `gitmodules_write:OK` — the literal pre-fix vulnerability |
| `unaudited_roots()` returns empty | e2e audit test fails, stderr showing exactly `no project file changes` for a session that rewrote the sibling |
| `unaudited_roots` filter polarity flipped | unit test fails |

The two #212 mutations fail *different* tests, so the path-shaped denies and the resolver are independently proven.

The new e2e tests each carry a `write_control` assertion proving the granted root **is** writable, so a failure means the deny fired rather than the root being unreachable.

lib 728, doc 26, e2e 138 (+6 pre-existing ignored), e2e_guards 113, e2e_projects 46, integration 54, unit_tests 274 — 0 failures. fmt and clippy clean.

## Merge ordering — matters

This branch adds a **new parent-side git invocation** (`git rev-parse --git-common-dir`, one per grant, with cwd inside a repo the agent can influence). #211 hardens every parent-side git call against repo-controlled config that names a program to execute.

`rev-parse` was not among the subcommands #211 measured as executing such a program — but #211 did not probe it explicitly either. **#211 should land first**, and this call should then be routed through `src/git.rs`'s choke point; there is a comment on `extra_git_dirs` saying so. If this lands first, that follow-up is required rather than optional.

## Not closed, and worth pushing back on

**#212's third scope bullet remains open.** `--project-dir ~/code` or `--allow-write ~/code` over a directory *containing* repos still leaves every child repo's hooks writable — the granted path itself is resolved, never repos nested under it. This is the same pre-existing gap the project root has.

A one-line macOS fix exists — a global `(deny file-write* (regex #"/\.git/hooks/"))` — and was deliberately not taken: it has no Bubblewrap equivalent, so it would widen the macOS/Linux divergence, and its blast radius (any tool writing hooks anywhere) could not be tested here. This is the biggest thing to challenge.

## Other things flagged rather than smoothed over

- **Linux never compiled.** crates.io is unreachable from the build sandbox, so `cargo check --target x86_64-unknown-linux-gnu` was impossible. The Linux-only code was hand-checked and two uncertain coercions verified with standalone `rustc` probes. CI is the real verification.
- **New breakage on macOS, by design:** `git config user.email` and `git init` now fail inside a *granted* repo, as they already did in the project. Parity — but someone's workflow may depend on it.
- `git_common_dir`'s internal worktree test changed from a raw-string `== ".git"` comparison to a canonicalized path comparison. Equivalence was argued for all shapes and every existing test passes, but it is a shared function and the change has no dedicated regression test.
- #214 names non-repo grants too (a relocated `GOPATH` landing in `allow_write`, say). Completeness was chosen over quiet: filtering to repos would imply everything else *was* covered. It may read as noise.
- #214 residual: a grant pointing at a git-ignored subdirectory *inside* `project_dir` is still silently unaudited — `git status` does not report it and `unaudited_roots` cannot tell.
- The issue's open question — whether `--allow-write` at a git repo should require explicit opt-in — is not acted on here.

## Second pass — adversarial review

**Verdict after fixes: the mechanism was kernel-confirmed correct across every repo shape** — normal, inside-repo, bare, worktree, submodule, symlinked grant — and nothing was blocked. Three things closed since.

**The security-critical deny loop had zero host-independent coverage.** Deleting the entire `for dir in extra_git_dirs` block — which kills hook and config denies for bare repos, worktrees, submodules and inside-repo grants on macOS — left **all 728 lib + 54 unit + 274 integration tests passing.** Only the seatbelt-gated e2e tests caught it, and those never run on Linux CI.

That is the same shape as the wrong-anchor mutant in #203 that left 712/712 green. `profile_denies_resolved_git_dirs_of_granted_repos` now feeds `generate_profile` a non-empty `extra_git_dirs` and asserts all four rules at byte offsets **after** the corresponding user allow, so last-match-wins is pinned rather than mere presence. Pure string generation, no `require_sandbox!`, runs on Linux CI. Mutation-checked against the exact deletion.

**`docs/security.md` submodule wording corrected, and re-verified rather than assumed.** It said `.git/modules/<name>/hooks` are not covered. With real parent/child repos: a submodule granted **directly** via `allow.write` has its gitlink resolved into the parent's `modules/<name>`, which then gets the denies. It stays uncovered only when merely nested inside a granted parent. Both cases now distinguished.

**Also cleared, so it is not re-litigated:** the `git_common_dir` refactor produces no new grant in the dangerous direction — all four shapes match `main`'s semantics, and the canonicalized comparison cannot return `Some` where the raw `== ".git"` did not. Deny/allow ordering holds against the kernel even with an overlapping `--allow-write sibling` and `--allow-read sibling/.git`. `unaudited_roots` names the right set across six cases, including a sibling sharing a string prefix.

**And the #211 ordering dependency is softer than stated above:** `git rev-parse --git-common-dir` was probed in an attacker-controlled repo against `core.fsmonitor`, `core.alternateRefsCommand`, `core.pager`, `core.sshCommand`, `uploadpack.packObjectsHook` and `diff.external` — no markers written, no execution. Still route it through #211's choke point when that lands, but it is not a hole today.

## One consequence worth a conscious decision

On Linux, Bubblewrap leaves `.git/config` writable by a pre-existing deliberate decision — read-only binding it breaks `git config` / `remote` / `submodule` and risks a stale `config.lock`. **That decision was made for one repo. It now applies to N.**

A `core.hooksPath` redirect therefore survives in *every* granted sibling, not just the project; the hooks ro-bind only stops the direct file-planting vector. macOS has no such residual — it denies `.git/config` for every writable root.

So a platform gap that was one-repo-wide before this change is now as wide as the grant list. That is not a regression introduced here — the alternative is not granting the sibling at all — but it is the kind of widening that should be nodded at rather than discovered later.

unit_tests 275 (was 274). Everything else unchanged, 0 failures.


Starefossen:
Rebased onto main, which now carries the git-persistence hardening from GHSA-g79x-v9p4-jmxw. The two branches overlapped in all three files, so this is worth a re-read rather than a re-approve.

**How the conflicts were resolved.** Main restructured every gitdir rule into `emit_gitdir_denies`, driven by `emit_git_persistence_denies` emitted last in the profile. This PR had looped the older two-rule form over writable roots. Main's structure won, and this PR's coverage is re-applied as inputs to it: the emitter now builds a deduplicated gitdir list from every writable root, plus `git_common_dir`, plus every `extra_git_dirs` entry, and runs main's full deny set over each.

**One judgment call, stated rather than buried.** That gives granted roots more than this PR originally asked for. They now also get the `file-write-unlink` and `file-write-data` pins, the `commondir` deny, and the `modules` subtree deny, not just hooks and config. Uniformity was chosen over the narrower loop because per-gitdir divergence is how the rename bypass got in. Visible cost: `rm -rf` or `mv` of a granted repo's `.git` is now blocked, and so is removing a granted bare repo's directory. If that is the wrong trade, the fix is a flag on `emit_gitdir_denies`, not a revert.

**Two smaller ones.** `git_dir_of` now routes through `crate::git::command`, which its own comment had asked for once #211 landed. And main's `raw == ".git"` early return is gone, subsumed by comparing against the canonicalized `<project>/.git`, which also catches absolute spellings and a `../..` that resolves back.

**Verification.** 1418 tests pass, fmt and clippy clean. Every protective line was mutation-tested: delete it, confirm a test fails, restore. That includes this PR's own two, the writable-root denies and the audit fix.

One gap was found and closed rather than reported. The writable-roots deny was caught only by a `require_sandbox!` e2e test, which never runs on Linux CI, so on Linux that mutation went green. Host-independent assertions were added to `profile_denies_resolved_git_dirs_of_granted_repos`, whose commit message already claimed that coverage. I re-ran the mutation myself against the pushed branch to confirm: it fails with the loop neutered and passes with it restored.

Kernel-verified on macOS. Rename, hook write, config append, commondir write, and rmdir all blocked on the project repo. With `--allow-write` on a sibling, the sibling's `.git/hooks` and `.git/config` are blocked while an ordinary write to the sibling succeeds, so the deny is what fired rather than the grant being absent.

Last commit is a comment-only fix. `git_dir_of` argued it needed no path filtering because its result "can never widen access". True of a path it returns, wrong about a `None`, which drops the denies entirely and so fails open. That is safe today only because `rev-parse` sits on `CONTENT_FREE_SUBCOMMANDS`; take it off and a hostile `filter.*.clean` in a granted repo would silently disarm that repo's hook deny.

PR #268, opened 2026-09-03
fix(macOS): allow Playwright Chrome for Testing startup

### LINKED ISSUE #263: macOS: Chrome for Testing fails despite Playwright cache exception
## Problem

On macOS, Chrome for Testing cannot start inside cplt even when the documented Playwright cache exception is enabled. This prevents Playwright MCP from performing browser operations: the MCP server can initialize and list tools, but launching the browser fails.

Verified on:

- macOS 26.6.2, Apple Silicon
- cplt `2026.08.24-153138-0d1d66d`
- latest released cplt `2026.09.02-225851-d840db9`
- Playwright Chrome for Testing revisions 1217 and 1243
- `@playwright/mcp@0.0.80`

## Minimal reproduction

With a Playwright-managed Chrome for Testing already installed, set `CHROME` to its executable. For example:

```bash
CHROME="$(find "$HOME/Library/Caches/ms-playwright" -path '*/Google Chrome for Testing.app/Contents/MacOS/Google Chrome for Testing' -type f | head -1)"
mkdir -p .tmp-chrome-direct .tmp-chrome-cplt
```

The browser starts successfully without cplt:

```bash
"$CHROME" \
  --headless \
  --disable-gpu \
  --no-first-run \
  --user-data-dir="$PWD/.tmp-chrome-direct" \
  --dump-dom about:blank
```

The equivalent invocation through cplt fails:

```bash
cplt \
  -d "$PWD" \
  --allow-cache-exec ms-playwright \
  -y \
  exec -- \
  "$CHROME" \
  --headless \
  --disable-gpu \
  --no-first-run \
  --user-data-dir="$PWD/.tmp-chrome-cplt" \
  --dump-dom about:blank
```

Observed fatal error:

```text
bootstrap_check_in com.google.chrome.for.testing.MachPortRendezvousServer.<pid>: Permission denied (1100)
```

Crashpad also reports denied extended-attribute operations before the fatal error.

## Evidence

The generated cplt profile currently permits Mach registration only for the `org.chromium.*` namespace:

```scheme
(allow mach-register (global-name-regex #"^org\.chromium\..+$"))
```

Chrome for Testing instead registers `com.google.chrome.for.testing.MachPortRendezvousServer.*`. The cplt source comment correctly explains that a missing `mach-register` permission causes `EPERM (1100)` during Chromium startup, but the allowed namespace does not cover the browser Playwright currently installs and selects.

`--allow-cache-exec ms-playwright` does successfully permit execution from the Playwright cache. The remaining failure is therefore a separate macOS runtime-policy gap rather than a cache-execution problem.

This was also reproduced through `@playwright/mcp@0.0.80`: initialization and `tools/list` succeeded, while a model-free `browser_navigate` to `about:blank` failed at browser launch with the same Mach-service denial.

## Expected behavior

When `ms-playwright` is explicitly enabled through `allow_cache_exec`, a Playwright-managed Chrome for Testing should be able to start under the macOS sandbox. Without that opt-in, the additional browser runtime permissions should remain unavailable.

## Suggested acceptance criteria

- Chrome for Testing can perform a model-free headless navigation to `about:blank` through cplt on macOS when `ms-playwright` is explicitly enabled.
- The necessary Mach registration permission is narrowly scoped to the browser service namespace rather than permitting arbitrary Mach-service registration.
- A macOS integration test launches the real browser through `sandbox-exec`; a generated-profile string assertion alone is not sufficient.
- The full launch is retested after the Mach denial is fixed to determine whether the observed Crashpad xattr denials need an additional scoped rule.

Related context: navikt/copilot#601 and navikt/copilot#625.

AudunSorheim:
## What

- make Playwright-managed Chrome for Testing usable inside cplt on macOS when `allow_cache_exec` explicitly enables `ms-playwright`
- allow only the Chrome for Testing Mach-service namespaces required at runtime
- create a short, random per-session `PWTEST_SOCKETS_DIR` and grant Unix-socket operations only beneath that concrete directory
- set `PLAYWRIGHT_MCP_SANDBOX=false` for opted-in children, unless the caller explicitly passes the variable through
- require functional Chrome navigation and live DOM evaluation in the macOS integration test

## Root cause

Browser startup failed at four successive runtime boundaries:

1. Chrome for Testing could not register `com.google.chrome.for.testing.MachPortRendezvousServer.<pid>`.
2. Its helpers then could not register `com.google.chrome.for.testing.apps.<64 uppercase hex>`.
3. Chromium attempted to initialize nested Seatbelt profiles and macOS rejected this with `forbidden-sandbox-reinit`.
4. Playwright MCP's control server exceeded the Unix-socket path limit under cplt's long `TMPDIR`, then received `EPERM` when binding under a shorter but unauthorized path.

The earlier DevTools-only test became green before renderer, GPU and network-service startup completed. The replacement test performs a real CDP navigation to a deterministic `data:` page, evaluates the live DOM, checks browser liveness and performs bounded process-group cleanup.

## Security boundary

All browser-specific behavior remains conditional on an exact `ms-playwright` cache-exec entry or a valid subpath such as `ms-playwright/chromium-1243`. Missing opt-in, unrelated entries, near-miss names and `allow_cache_exec_any` do not activate it.

The Mach rules are fully anchored to:

```text
org.chromium.*
com.google.chrome.for.testing.MachPortRendezvousServer.<numeric pid>
com.google.chrome.for.testing.apps.<exactly 64 uppercase hexadecimal characters>
```

For Playwright control sockets, cplt creates `/private/tmp/cplt-pw-<32 lowercase hex>` as a current-user-owned, non-symlink `0700` directory. The generated profile grants only `network-bind`, `network-inbound` and `network-outbound` for Unix sockets beneath that exact directory. It adds no general `/private/tmp` socket capability, temp execution, file-map-executable or xattr permission. Explicit `--pass-env PWTEST_SOCKETS_DIR` suppresses both the automatic directory and its SBPL rules; cplt now warns when that override has no value.

Chromium cannot initialize its own sandbox inside cplt's macOS Seatbelt profile or Linux seccomp policy. Under the explicit Playwright opt-in, cplt therefore sets `PLAYWRIGHT_MCP_SANDBOX=false`. An explicit `--pass-env PLAYWRIGHT_MCP_SANDBOX` returns the choice to the caller, and repository `deny.env` still applies last.

cplt remains the outer kernel boundary, but a compromised renderer receives the complete cplt Playwright profile rather than Chromium's narrower child profile. This trade-off is documented in `SECURITY.md`. Keeping the setting inside cplt avoids disabling Chromium sandboxing for users of the shared MCP registry recipe who do not run through cplt.

Crashpad xattr denials remain observable but are not functional blockers, so no xattr permission is added.

## Verification

- formatting and Clippy pass on the merged branch
- 298 unit tests pass
- all focused Playwright runtime, environment and policy tests pass
- the complete integration suite passes: 61 tests
- 49 e2e project tests pass
- the broader sandbox-safe suite reaches 797 passing library tests and 144 passing e2e tests; the remaining 3 proxy and 5 network/Docker probe failures are caused by this session already running inside cplt
- the kernel-enforced socket test covers bind, inbound and outbound access beneath the generated directory while denying a sibling
- the functional Chrome test performs CDP navigation and live DOM evaluation
- model-free acceptance with `@playwright/mcp@0.0.80`, Chromium revision 1243 and `browser_navigate("about:blank")` succeeds and exposes 24 tools

## Known limitation and rollback

Normal exits remove the per-session socket directory. `SIGKILL` can leave an empty random `0700` directory behind. Automatic collection is intentionally out of scope because a safe implementation requires descriptor-relative traversal and bounded depth, breadth and runtime.

Rollback is a revert of this PR. Users without explicit Playwright cache execution retain existing behavior.

Fixes #263

Runtime support here is required by navikt/copilot#625 and navikt/copilot#601.


Starefossen:
Reviewed this adversarially. **Verdict: merge.** Nothing lands outside the `ms-playwright` opt-in, and it supersedes the narrower fix from #272 cleanly rather than duplicating it: the generated profile has exactly three `mach-register` lines, not four.

I pushed one commit to the branch (`ac1cf5c`) restoring an assertion that #272 had added and this PR replaced. `chromium_runtime_rules_absent_by_default` now checks both the two specific rules *and* the blanket `!p.contains("mach-register")`. The specific checks say what is expected; the blanket one is the tripwire that catches a namespace someone adds later without the gate. It passes as-is, so this is belt-and-braces rather than a fix.

## Verified

- **Gating.** Generated profiles against an empty config: default, `ms-playwright-evil`, and `--allow-cache-exec-any` all give 0 `mach-register`, 0 `cplt-pw`, 0 `syscall*`. `ms-playwright` and `ms-playwright/chromium-1243` give 3/3/1. The child env has neither variable without the opt-in.
- **Socket directory.** 128 bits from `/dev/urandom`, `DirBuilder.mode(0700).create` is atomic, a collision errors rather than adopting an existing directory, and `prepare()` re-stats for real-dir, uid, mode `0700` and `canonicalize == path`. The grant is a `subpath` of that exact directory; the sibling-bind deny is covered in-kernel by a test.
- **Regexes.** Anchored, and each requires content after the trailing dot, so `com.google.chrome.for.testingX` cannot match.
- **Ordering.** Socket allows at profile lines 288-290, mach-register at 53-55. Everything emitted later is `file-*` or tcp/ip, so nothing reopens or revokes them.

## Two things worth a follow-up, neither blocking

**`--pass-env PWTEST_SOCKETS_DIR` does not work.** Passing the key suppresses both the directory creation and the SBPL grant, so the child gets the caller's path and `bind()` fails with `EPERM`; passing it with no value only warns. SECURITY.md describes this as preserving the caller's value, which reads as intentional, so I have not touched it. If the override is meant to work it needs the grant to follow the caller's path; if it is not, rejecting the key with an error would be clearer than a silent `EPERM`.

**SIGINT leaves the socket directory behind.** Only SIGTERM and SIGHUP are forwarded (`sandbox_exec.rs:631-640`), so Ctrl-C leaves an empty `0700` `/private/tmp/cplt-pw-<id>`. The docs mention SIGKILL but not this. It carries no capability once the process is gone and there is no `gc_stale` equivalent for it — housekeeping rather than a security issue.

One judgement call I want to flag rather than decide: narrowing `com.google.chrome.for.testing\..+$` to two suffix-exact patterns buys little, since any registrant of that namespace is already inside a profile holding `syscall*`, AF_UNIX and IOKit. The suffix shapes were observed on one alpha build, and a future build registering a different suffix fails with EPERM 1100 and segfaults. Keeping the `.+$` form would be more durable. Your call — you have run this against the real browser and I have not.

Issue #246, opened 2026-09-02
feat: allow.exec / --allow-exec for user-granted executable trees


Starefossen:
Two open reports converge on the same missing capability, and the workaround offered in #152 does not work.

- **#202** asks for it directly ("Add allow-exec configuration option"). #232 fixed the documented Homebrew-on-Linux prefix, but a source-built Homebrew under `~/.linuxbrew` — or any relocated `HOMEBREW_PREFIX` — is still broken with no workaround.
- **#152** is a relocated `CARGO_HOME`. `CARGO_HOME` is in `TOOL_PATH_ENV_VARS` with `write: true`, so `merge_tool_path_env_overrides` pushes the relocated directory into `resolved.allow_write`, and the Linux rule for `extra_write` sets `execute: false` (`sandbox_landlock.rs:548-557`). `$CARGO_HOME/bin/*` is then EACCES.

**Correction to the workaround suggested on #152:** `allow.read = ["<CARGO_HOME>/bin"]` does not grant exec on Linux — `extra_read` sets `execute: false` (`sandbox_landlock.rs:535-545`), verified against a real kernel. On macOS the bug does not reproduce at all, because `process-exec` is granted globally (`sandbox_profile.rs:246`). The reporter is on Linux, which is the one platform where the suggestion fails. The root-cause description there is also off: Landlock is additive, so nothing "denies" the exec — the write grant simply never carries it, and there is no ordering to get wrong (the #158 trap is macOS-only).

There is currently **no opt-in exec grant on either platform**: `extra_read` is `execute: false` on Landlock and `file-read*` only in SBPL (`sandbox_profile.rs:1281-1284`). So "make it opt-in" was never available without building this.

## Shape

`allow.exec` in config plus `--allow-exec <path>`, threaded through `Resolved`, `SandboxConfig` and both backends. On Landlock the rule needs read+execute (exec without read is EACCES — the ELF interpreter must be readable too). On macOS `process-exec` is already blanket, so the work is `file-read*` plus `file-map-executable` and, critically, a **write-deny emitted last** so an exec grant cannot combine with a write grant on the same tree. That ordering is exactly where #158 went wrong.

[...]

Issue #240, opened 2026-09-02
Linux: pathname Unix sockets are not mediated, so the session manager is reachable


Starefossen:
Landlock cannot restrict `connect()` to a pathname Unix socket before ABI v9 (`LANDLOCK_ACCESS_FS_RESOLVE_UNIX`, kernel 7.1). cplt hard-requires V1 and best-efforts V6 (`sandbox_landlock.rs:945-955`), clamping the probed ABI at V6 (`:898-905`), and handles only `AccessNet::ConnectTcp` (`:961-963`). The seccomp filter blocks no socket syscall (`:1194-1228`). Bubblewrap does `--ro-bind / /` with no `--unshare-net`, and a read-only bind does not stop `connect()` — the ro flag is checked by `mnt_want_write()`, while unix connect only checks inode DAC write.

Consequences, all reachable with default settings:

- `systemd-run --user -- sh -c '...'` via `/run/user/$UID/bus` (or systemd's private socket) asks the user service manager to spawn a process. The manager is not a descendant of the sandbox, so nothing is inherited — full escape, no group membership required, one command. Reachable **with and without** bubblewrap.
- The Docker socket (`/var/run/docker.sock`, or rootless under `/run/user/$UID`) for anyone in the `docker` group — host root via a bind mount.
- Podman's user socket where that unit is active.
- tmux `run-shell` via `/tmp/tmux-$UID/default`, when bwrap is absent or the socket lives outside `/tmp`.
- Abstract sockets (X11 and similar) — `LANDLOCK_SCOPE_ABSTRACT_UNIX_SOCKET` (ABI v6) is available on kernel ≥ 6.12 but is not requested either.

`TMUX`, `XDG_RUNTIME_DIR`, `DOCKER_HOST` and `CONTAINER_HOST` are all passed through in default sanitized mode (`sandbox_policy.rs:172,198,225,226`), but that only saves the attacker a guess — the paths are deterministic.

This also means `--allow-socket` and the gpg-agent socket rules (`sandbox_landlock.rs:561-571`, `:587-597`) are **no-ops on Linux in both directions**: they emit read/write `FsRule`s, which `connect()` never consults. They grant nothing and deny nothing.

## What can and cannot be done

[...]

Issue #207, opened 2026-08-27
DENIED_FILES is overridable via allow.read on Linux, contradicting the documented hard-deny guarantee


Starefossen:
`DENIED_FILES` is documented as the hard-deny list: `~/.netrc`, `~/.pypirc`, `~/.gem/credentials`, `~/.vault-token` and friends, "not overridable" — as distinct from `DENIED_HOME_SUBPATHS`, which is deliberately overridable with `allow.read`.

On Linux that distinction does not exist. `cplt config set allow.read "~/.netrc"` produces a working read grant.

## Why

On Linux, `DENIED_FILES` entries are deny-by-**omission**: nothing emits a deny rule for them, they are simply never granted. The repo's own test asserts exactly this (`src/sandbox_landlock.rs:1640`):

```rust
fn denied_files_not_in_ruleset() {
    ...
    for file in policy::DENIED_FILES {
        let path = home.join(file);
        let found = policy.fs_rules.iter().any(|r| r.path == path);
        assert!(!found, "denied file {file} should NOT be in ruleset");
    }
}
```

Meanwhile `config.extra_read` is pushed into the Landlock ruleset unfiltered (`src/sandbox_landlock.rs:535`):

```rust
// ── Extra read paths from config ──
for p in config.extra_read {
    fs_rules.push(FsRule {
        path: p.clone(),
        access: FsAccess { read: true, write: false, ... },
```

Nothing between config parsing and ruleset construction validates `allow.read` against `DENIED_FILES`. Landlock is grant-only, so an added grant simply wins — there is no deny for it to lose to.

On macOS the guarantee is real: generic `extra_read` allows are emitted before the literal denies, and only `DENIED_HOME_SUBPATHS` / `DENIED_DOTFILES` get post-deny re-allows (`src/sandbox_profile.rs:1151`). A user cannot re-allow a `DENIED_FILES` entry there.

## Why it matters

The two lists exist precisely to express "overridable" versus "not". On Linux they collapse into one, so a security property the docs state as absolute holds on one platform only — and silently, since the user gets no warning that their `allow.read` did something the tool documents as impossible.

This is the same shape as several bugs fixed recently: the rule is present in the model and absent in the kernel, and nothing tells anyone.

## Fix

[...]

PR #312, opened 2026-09-04
fix(sandbox): honour write grants in credential dirs, refuse the dirs themselves

### LINKED ISSUE #291: Linux grants ~/.ssh at directory level where macOS denies it
Split out of #207, which closed the `allow.read` hole for `DENIED_FILES` but deliberately left this one alone.

`DENIED_DOTFILES` entries diverge at the directory level. `allow.read = "~/.ssh"` is granted on Linux; on macOS the deny wins and the grant does nothing. Files *inside* those directories are a different case and not a divergence — they are overridable by design on both platforms, via `emit_denied_dotfile_overrides` on macOS.

## Why this is not a quick fix

`~/.ssh` is a supported grant on Linux today. `brief.rs` actively steers users towards it when they need SSH keys. Closing the divergence by making Linux match macOS would break a workflow the tool currently recommends, so it is a behaviour change and not a bug fix, whichever way it is resolved.

## The three ways out

1. Make Linux match macOS: deny the directory, keep per-file overrides. Consistent, and it breaks the documented SSH workflow until something replaces it.
2. Make macOS match Linux: allow the directory grant on both. Consistent the other way, and it widens macOS.
3. Keep the divergence and document it as deliberate, with `brief.rs` telling Linux and macOS users different things.

Option 1 is the one that matches the rest of the security model, but it needs an answer for the SSH case first — that is the actual work here, not the deny rule.

## Related

#207 fixed the `DENIED_FILES` half and documented the remaining ancestor-grant gap on Linux (`allow.read = "~/.gem"` still exposes `~/.gem/credentials`, because Landlock cannot deny a subpath inside an allowed directory). This issue is the `DENIED_DOTFILES` directory-level case, which is separate: it is not a kernel limitation, it is a policy choice nobody has made explicitly.

Starefossen:
Closes #291, and fixes a second defect of the same shape found while investigating it.

Both halves are the pattern this repo keeps finding: a grant that is silently inert on one platform and effective on the other. #207 made a silently-dropped grant a startup error, #246 refuses an overlap rather than half-honouring it, #240 found `--allow-socket` inert on Linux while looking present. These are two more.

## Half 1 — `allow.write` inside a credential directory was inert on macOS

`emit_denied_dotfile_overrides` consulted `extra_read` only. A write grant on a file inside a `DENIED_DOTFILES` directory was emitted by `emit_user_allows` — before `emit_deny_rules` — lost to the subpath deny, and did nothing at all on macOS, not even read. On Linux the same grant worked. Nothing warned.

The concrete casualty is `~/.ssh/known_hosts`. Under the default `StrictHostKeyChecking`, stock `ssh` appends to it on a first connection, so the per-file SSH route — the one `brief.rs` recommends to users on both platforms — could not complete on macOS. There was no grant that permitted it.

Write grants now get a post-deny re-allow, with read alongside them, matching how `emit_user_allows` pairs the two. Scoped to the file named; siblings stay denied.

## Half 2 — a grant on the directory itself is refused (#291)

`allow.read = "~/.ssh"` was granted on Linux and ignored on macOS. Neither platform told the user which they were getting. It is now a startup error on all four grant keys, and the Landlock backend filters it too, so the ruleset invariant holds for every caller of `generate_policy` and not only the ones that went through `prepare`.

```
[cplt] allow.read names /Users/hans/.ssh (~/.ssh), a credential directory that cannot be
granted whole: macOS denies it whatever the grant says, so honouring it on Linux alone would
mean the same config opens every key in it on one platform and nothing on the other. Name the
specific files you need instead, e.g. ~/.ssh/<file> — that grant works on both.
```

This is option 1 from the issue. Its stated cost — "it breaks a workflow the tool currently recommends" — turned out not to hold on current `main`: nothing recommends the directory. `brief.rs:135` already tells users on both platforms to ask for "an `allow.read` of the specific key file", and no doc suggests `~/.ssh`. Half 1 is what makes that per-file route actually complete, which is why it is in the same PR and why it was built first — closing the directory route while the per-file route could not write would have closed the only working door.

An error rather than a silent narrowing, so the Linux behaviour change arrives as a message carrying its own fix.

`brief.rs` loses the `cfg!(macos)` branches that existed only to model this divergence. One divergence remains there, retargeted at what actually causes it.

## Deliberately not in scope

- **An ancestor grant (`allow.read = "~"`) is not refused.** It still exposes everything under it on Linux, because Landlock cannot deny a subpath inside an allowed directory — the same limitation #207 documents. `brief.rs` still reports it honestly on Linux, and `docs/known-impacts.md` names it as a remaining gap.
- **`--allow-docker`'s read-only `~/.docker` grant is exempt** and must stay so: it is a first-party rule the backends emit, not a user grant on that path, so the refusal cannot see it. `prepare_does_not_refuse_the_allow_docker_dotfile_grant` pins that — without it the flag would refuse every run it is set on.

## Evidence — what was run and what was read

**From a live sandbox probe on macOS**, using `cplt check path` (live probe plus static model) and `cplt exec`:

- Before: `--allow-read ~/.ssh` blocked the directory *and* every file in it; `--allow-write ~/.ssh/<file>` reported `read: blocked, write: blocked`, both with `live probe says BLOCKED, the policy model predicted ALLOWED`.
- After: the write grant reports `read: allowed, write: allowed` with no mismatch, a real `>> ~/.ssh/<file>` append succeeds inside the sandbox, and a sibling file in the same directory stays denied.
- After: the directory grant is refused on `--allow-read`, `--allow-write`, `--allow-exec` and `--allow-socket`, and `--allow-docker` still launches.

**From reading the code, not run here:** everything about Linux. This checkout runs on macOS, so the Linux half rests on `sandbox_landlock.rs` and on the cross-platform `generate_policy` model — which is exactly what `check path` was reporting as "the policy model predicted ALLOWED" on a macOS host. The `SSH_AUTH_SOCK` findings in the issue thread are likewise read-only: no agent was running in the test shell.

## Mutation checks

Both defects were reintroduced and the specific test watched to fail, then restored:

- Passing `&[]` for `extra_write` at the `emit_denied_dotfile_overrides` call site → `profile_extra_write_overrides_a_denied_dotfile_file` fails on `missing post-deny re-allow: (allow file-read* (literal ".../known_hosts"))`.
- Dropping `denied_dotfile_dir` from `grant_is_refused` and from `validate_hard_denied_grants` → `prepare_rejects_a_grant_on_a_denied_dotfile_directory` and `a_granted_denied_dotfile_directory_stays_out_of_the_ruleset` both fail.

## Gates

`cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, `cargo test --lib` (883), `cargo test --test unit_tests` (318), `cargo test --test integration` (64) — all clean on the rebased branch, after #300 cleared the clippy backlog.

### DIFF
--- a/docs/known-impacts.md
+++ b/docs/known-impacts.md
@@ -215,6 +215,22 @@ Bubblewrap's private `/tmp` helps only for that stock layout. The desktop agents
 
 If your keys must be unusable by a compromised agent on Linux, unload them (`ssh-add -D`) before starting the session.
 
+### If you do need SSH inside the sandbox
+
+Three things have to line up, and the same configuration works on both platforms:
+
+```bash
+cplt config set allow.ports 22                      # outbound is 443-only by default
+cplt config set allow.read  "~/.ssh/id_ed25519"     # the key, by name
+cplt config set allow.write "~/.ssh/known_hosts"    # ssh appends on a first connection
+```
+
+Port 22 is the first gate: without it no key grant matters, because the connection never opens. `~/.ssh/config` needs its own `allow.read` if you have one. `known_hosts` needs **write**, not read — under the default `StrictHostKeyChecking` a first connection to a host appends to it, and a read-only grant fails the connection.
+
+**The directory itself cannot be granted.** `allow.read = "~/.ssh"` (or `~/.aws`, `~/.gnupg`, any of the credential directories) is refused at startup with an error naming this per-file route. macOS never honoured such a grant — the blanket subpath deny beats it whatever the config says — while Linux granted it for real, so the same config file opened every key on one platform and nothing on the other. Refusing is the only answer both backends give alike.
+
+One gap remains on Linux, unchanged and the same limitation described under [private registries](#private-registries): a grant on a *parent* — `$HOME` itself — still exposes everything under it, because Landlock cannot deny a subpath inside an allowed directory.
+
 ## D-Bus and systemd (Linux) — the session manager is reachable
 
 `XDG_RUNTIME_DIR` is on the environment allowlist, and Landlock cannot restrict `connect()` to a pathname unix socket before ABI v9 (kernel 7.1), so on a current kernel `$XDG_RUNTIME_DIR/bus` is reachable from inside the sandbox unless bubblewrap masks it. This is established by reading the code, not by running it on a host — treat it as reachable rather than demonstrated. `systemd-run --user <cmd>` hands the command to the user's systemd instance, which starts it as a new unit **outside** Landlock, seccomp and the bubblewrap namespaces. The same applies to any other D-Bus service on the session bus that can start a process.
--- a/src/brief.rs
+++ b/src/brief.rs
@@ -163,10 +163,10 @@ pub fn generate_session_brief(resolved: &Resolved, agent: Agent, home: &Path) ->
          and similar) are unreadable by design (EPERM). Don't retry; tell the \
          user.\n",
     );
-    // `allow.read` of a path inside one of those directories is a supported
-    // configuration — the backends re-allow it after the blanket deny — so the
-    // blanket claim above needs its exceptions spelled out, or the agent will
-    // report a working read as blocked.
+    // `allow.read` or `allow.write` of a path inside one of those directories is
+    // a supported configuration — the backends re-allow it after the blanket
+    // deny — so the blanket claim above needs its exceptions spelled out, or the
+    // agent will report a working access as blocked.
     let credential_overrides = denied_dotfile_overrides(home, &resolved.allow_read);
     if !credential_overrides.is_empty() {
         let _ = writeln!(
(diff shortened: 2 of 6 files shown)


Starefossen:
Follow-up commit `616f330`: the first push failed `brief_treats_a_bare_ssh_dir_grant_as_no_grant` on Linux CI (unit-tests and linux-integration-tests), and only there.

My simplification of `ssh_key_readable` and `denied_dotfile_overrides` excluded the directory itself inside the descendant test but not inside the Linux ancestor test, and `~/.ssh.starts_with(~/.ssh)` is true — so a bare directory grant still read as a real grant on Linux, which is the one thing this PR is removing. Fixed by excluding the directory once, ahead of both prefix tests.

Worth recording, because it is the same class of thing this PR is about: the macOS checkout could not see it. The Linux branch of both predicates is verified by CI only, not by anything I ran locally. The cross-platform test that now covers it is the one that caught it.

Both suites green on the rebased branch.

Issue #152, opened 2026-07-15
CARGO_HOME is not read to figure out what to set an executable permission on


krageon:
CARGO_HOME appears to be passed to the executing agent, but it is not read by cplt to figure out where to set executable permissions. 

That means the agent you start will not be able to execute any binaries in the bin folder of cargo if CARGO_HOME is not the default (~/.cargo). You can test this by setting CARGO_HOME to something like ~/.local/share/cargo and attempting to run an agent. If you ask it whether it can execute a tool that is provided by cargo, you will get the result that the folder is mounted with noexec.

The workaround for this right now is to `ln -s "${CARGO_HOME}" ~/.cargo`, but that's not ideal if you want to have a homedir that is f.e. XDG spec compliant.

krageon:
As a broader comment it would be nice if I could tell the tool which folders to mount as executable. Right now I can only make specific predefined trees (such as a cache or tmp folder) executable. There isn't something like `--allow-exec <folder>` as far as I can tell. That would have allowed for a more elegant workaround for this bug.

Starefossen:
Confirmed, and the root cause is narrower than "CARGO_HOME is not read" — it *is* read, but the relocated path loses the exec semantics the default has.

`CARGO_HOME` is in `TOOL_PATH_ENV_VARS` (`sandbox_policy.rs`) with `write: true`, so `merge_tool_path_env_overrides` pushes the relocated dir into `resolved.allow_write`. But the default `~/.cargo` isn't granted as one writable blob — it's three `HOME_TOOL_DIRS` entries with deliberately different posture:

| path | process_exec | map_exec | write |
|---|---|---|---|
| `.cargo/bin` | ✅ | ✅ | ❌ |
| `.cargo/registry` | ❌ | ✅ | ✅ |
| `.cargo/git` | ❌ | ✅ | ✅ |

A relocated `CARGO_HOME` collapses all three into a single write grant. Write-without-exec is precisely what the write-then-exec protection denies, so `$CARGO_HOME/bin/*` becomes non-executable — your "mounted with noexec" observation.

[...]

PR #319, opened 2026-09-04
fix(sandbox): keep write grants non-executable, stop map_exec meaning execve

### LINKED ISSUE #243: Write-then-exec protection does not hold for user grants (macOS) or tool dirs (Linux)
Two related posture bugs, both found while triaging #152, both meaning the write-then-exec protection does not hold where the docs imply it does.

## macOS: user `allow.write` grants are already rwx

`process-exec` is granted globally at `sandbox_profile.rs:246`, and `emit_user_allows` (`:1278-1291`) emits only `file-read*` and `file-write*` — no exec deny. Verified with `--print-profile --allow-write <dir>`: line 25 is `(allow process-exec)`, the grant appears at lines 328-329, and no deny is emitted.

So a tree the user grants write on is writable and executable. The curated `HOME_TOOL_DIRS` entries get a compensating deny when `write && !exec` (`:809-822`), but user-granted trees never do.

Consequence for #152: the relocated-`CARGO_HOME` bug does not reproduce on macOS at all, and it reproduces on Linux for the opposite reason — there the write grant simply never carries execute. Adding a `(deny process-exec)` for `extra_write` on macOS would make #152 reproduce there too, so it should not be done before the exec-grant work lands.

## Linux: `map_exec` is silently promoted to `process_exec`

`sandbox_landlock.rs:346` (and `:372`, `:648`) computes `execute: dir.process_exec || dir.map_exec`. Landlock has no separate right for `mmap(PROT_EXEC)` — libraries need only `READ_FILE` — so `map_exec` collapses into full execute permission.

The `HOME_TOOL_DIRS` entries that carry `map_exec` for dylib loading are therefore **write+execute** on Linux: `.cargo/registry`, `.cargo/git`, `.gradle`, `.m2`, `.nuget`, `.dotnet`, `go/pkg`.

The posture table in the #152 discussion describes macOS behaviour and does not hold on Linux. A fix would be to use `process_exec` alone for the Landlock `execute` bit, with a carve-out for the cases that genuinely need to map executable pages from those trees (`.gradle/jdks` is the obvious one) — that needs testing against real Gradle and .NET runs before it ships.

Both should be verified on a Linux host; the macOS half was verified locally with `--print-profile`.

Starefossen:
Closes #243. Both halves, with the Linux half stopping where the evidence stops.

## macOS: `allow.write` was also an exec grant

`process-exec` is granted profile-wide, and `emit_user_allows` emitted only `file-read*` and `file-write*`. So a tree the user granted write on was writable **and** executable — the binary-drop pair `validate_exec_grants` refuses when an `allow.exec` grant creates it the other way round (#295). `HOME_TOOL_DIRS` entries have carried the compensating `(deny process-exec)` for years; user grants never did.

`emit_user_write_exec_denies` now emits it, after every allow, because SBPL is last-match-wins. Three execute rights are re-issued afterwards **and only where they fall inside the granted tree** — an unconditional re-allow at the tail would outrank the `/private/tmp` exec deny for a project or scratch dir that lives there:

- the project directory (write+exec by design; an `allow.write` on `~/Repos` would otherwise stop the project running its own build output),
- the scratch dir (write+exec by design),
- the `EXEC_IN_WRITABLE` trees, which were already write+exec *before* the grant.

Everything else loses execute, tool directories included: `allow.write = ["~/.cargo"]` makes `~/.cargo/bin` writable, and a writable `~/.cargo/bin` is a trojan that outlives the session, so it does not also stay executable.

## Linux: `map_exec` was silently promoted to `execve()`

`sandbox_landlock.rs` computed `execute: dir.process_exec || dir.map_exec`. Landlock checks `EXECUTE` on `execve()` alone (`file_open` with `FMODE_EXEC`) and has no `mmap(PROT_EXEC)` hook — a dynamic loader gets there with `READ_FILE` — so `map_exec` could not be honoured as anything narrower than full execve. Every `write: true, map_exec: true` dependency store was write+execute **on Linux and nowhere else**.

The bit now follows `process_exec` alone. The same two lines existed for app dirs and agent dirs; both are changed too, and both are no-ops today (every `AppDir` with `map_exec` also has `process_exec` on the same kinds, and no `AgentDir` sets `map_exec`) — changed so the rule cannot drift when one of them stops being a no-op.

## The carve-outs, tree by tree

`EXEC_IN_WRITABLE` in `sandbox_policy.rs`, in the shape of `PROTECTED_IN_ROOT`: one entry per tree that is both writable and executable, each with a reason and a `LinuxCoverage`.

**Dropped exec, verified:** `.cargo/registry`, `.cargo/git`, `.m2`, `.nuget`, `go/pkg`, `Library/pnpm/store`, `.local/share/pnpm/store`, and `.gradle` itself. The evidence is macOS parity plus the Landlock semantics above: each of these is `process_exec: false` on macOS today and emits an explicit `(deny process-exec)` there, so cargo, Maven, Gradle, NuGet, Go and pnpm already run daily with `execve()` denied on exactly these trees. What they need from them is `dlopen`, which Landlock grants with `READ_FILE`.

**Kept exec, and why:**

| Tree | Verdict |
|---|---|
| `.gradle/jdks` | **Carve-out, justified.** Gradle provisions toolchain JDKs here and execs them. macOS grants exactly this (`emit_gradle_toolchain_exec`) and takes the write back at the tail of the profile (`emit_gradle_toolchain_write_deny`); Landlock cannot do the second half, so it stays write+execute there — recorded as a `Gap`, not closed with a bubblewrap bind that would be absent on hosts without user namespaces. |
| `.bun/install` | **Write+exec by design on both backends.** `bunx` runs packages out of the cache `bun install` writes. Recorded so the pair is reviewable rather than an absence. `.bun/bin`, the PATH-resolved half, is write-denied separately. |
| `.dotnet` | **Not settled. Left alone.** macOS re-grants exec through the configured `dotnet_root` for SDKs installed by `dotnet-install.sh`; Landlock has no counterpart rule, so dropping `EXECUTE` here would break that install shape with nothing to compensate. Could not run a real .NET build on a Linux host. |
| `.konan` | **Not settled. Left alone.** Kotlin/Native execs the LLVM toolchain it downloads into `~/.konan/dependencies`. macOS denies `process-exec` here, so K/N may already be broken there — that could not be demonstrated either way from a macOS checkout. |

The last two are stated gaps, not decisions: this is macOS with no Linux host, a wrong carve-out either reopens the hole or breaks every user of the tool, and neither failure is visible from here.

## No silent grants

Per AGENTS.md and #317, what is not closed is written down rather than left as an absence: `EXEC_IN_WRITABLE` entries carry `LinuxCoverage::Gap(reason)`, and SECURITY.md gets a "Landlock cannot express `map_exec`" section plus a "User-granted write trees" bullet. `docs/security.md` gets the Linux-limitations entry. README's `--allow-write` row now says the tree is not executable and points at `--allow-exec`.

## Relocated `CARGO_HOME` (#152, #280) is unaffected

Checked before building on it. `merge_tool_path_env_overrides` turns `CARGO_HOME`, `GOPATH` and `PNPM_HOME` into `ToolRoot`s that re-root the `HOME_TOOL_DIRS` entries — they never reach `allow.write`, so `$CARGO_HOME/bin` keeps its exec-only posture and `$PNPM_HOME` keeps its shims. The write-granting env vars that are *not* relocatable (`GOMODCACHE`, `GOCACHE`, `YARN_CACHE_FOLDER`, `PIP_CACHE_DIR`) are caches that never needed exec.

Also verified before starting: #295 shipped `allow.exec` **and** `validate_exec_grants`, which refuses an exec grant overlapping a writable tree in `sandbox::prepare` — platform-agnostic, so on both backends. The sequencing constraint the issue set is satisfied.

## Tests

New: `write_grant_is_not_also_an_exec_grant`, `write_exec_deny_comes_after_every_broad_exec_allow` (asserts placement against *every* exec allow in the profile, not one known culprit), `write_grant_keeps_the_documented_write_exec_trees`, `write_exec_carve_outs_are_scoped_to_the_granted_tree`, `map_exec_alone_does_not_grant_execute`, `exec_in_writable_trees_keep_execute`, and a kernel-level `real_profile_write_grant_is_not_executable` that drops a real binary in a granted tree under `$HOME` (not temp, which is exec-denied anyway and would pass for the wrong reason) and runs it under `sandbox-exec`.

Changed: `home_tool_dirs_permissions_match` now asserts `process_exec` rather than the union.

**Mutation check.** Deleting the one `(deny process-exec (subpath "{p}"))` line fails all four macOS unit tests and the integration test, which reports `EXIT:0` — the dropped binary ran. Restoring it passes. Reverting the Landlock bit to `process_exec || map_exec` fails `map_exec_alone_does_not_grant_execute` on `.cargo/registry`.

Gates: `cargo fmt`, `cargo test --lib` (912), `cargo test --test unit_tests` (322), `cargo test --test integration` (65), `cargo test --test e2e_projects` (49), `cargo clippy --all-targets -- -D warnings` all clean.


Starefossen:
Two follow-ups from this PR: #322 adds the launch warning for a write grant that shadows an exec tool dir (the loose end noted at merge), and #323 tracks the suspected `~/.konan` exec problem the `EXEC_IN_WRITABLE` entry records.

Issue #324, opened 2026-09-04
Linux: ~/.copilot is granted write+execute by default, for a reason that does not hold


Starefossen:
Found while fixing the Fable review findings in #321, and left out of that PR because it is a default grant rather than a cross-backend divergence.

`generate_policy` grants `~/.copilot` `read + write + execute` on Linux whenever the agent needs it (`src/sandbox_landlock.rs:764-772`). The comment gives the reason:

> Execute needed for `dlopen()` of native `.node` addons (keytar, pty, computer).

**That reason does not hold on Landlock**, and #243 established why earlier the same day. Landlock checks EXECUTE on `execve()` only — `file_open` with `FMODE_EXEC` — and has no `mmap(PROT_EXEC)` hook at all. Loading a shared object needs `READ_FILE`, which the rule already grants. So the execute bit buys nothing for `dlopen`.

What it does produce is a **writable and executable tree, granted by default**, in the directory the agent writes its own config to. That is the binary-drop pair `validate_exec_grants` refuses when a user asks for it explicitly, and that #238 and #243 spent the day removing everywhere else. Here it arrives without anyone asking.

`~/.copilot` is also not in `writable_trees`, so an `allow.exec` overlapping it is not refused either — the same gap #321 closed for `agent_dirs`, `git_common_dir` and `/dev/shm`.

## What to check before changing it

The claim to verify is that nothing under `~/.copilot` is genuinely `execve`d. Native `.node` addons are `dlopen`ed, not executed, so the stated reason is wrong — but Copilot may spawn something else from there that nobody wrote down. Worth an actual run on Linux with EXECUTE dropped, rather than a code reading, since the failure mode if we are wrong is that Copilot stops working for every Linux user.

The macOS side needs its own look: `process-exec` is granted globally there, so the equivalent question is whether the SBPL profile emits a compensating deny for this path the way it now does for user write grants (#243).

## Related

#243 (`map_exec` collapsing into full execve on Landlock, and the exec table), #238 (write+exec on PATH-resolved dirs), #321 (writable trees the overlap refusal did not know about).

Issue #323, opened 2026-09-04
Suspected: ~/.konan is process-exec denied, but Kotlin/Native execs its downloaded LLVM toolchain from there


Starefossen:
**Suspected bug from code reading, not an observed failure.** I have not run a Kotlin/Native build under cplt, and I could not from this machine. Filing it so the analysis is not lost, and because settling it is cheap for anyone who has a K/N project.

## The claim

`~/.konan` is a `HOME_TOOL_DIRS` entry with `process_exec: false, map_exec: true, write: true` ([`sandbox_policy.rs:1208`](https://github.com/navikt/cplt/blob/531d3af521826a418d96d6b7e8c21e30d79277aa/src/sandbox_policy.rs#L1208)). Because the tree is writable and not `process_exec`, macOS emits a compensating `(deny process-exec (subpath "~/.konan"))` — the deny is asserted directly by [`profile_konan_has_map_exec_only`](https://github.com/navikt/cplt/blob/531d3af521826a418d96d6b7e8c21e30d79277aa/tests/unit_tests.rs#L3080):

```
.konan should NOT have process-exec
.konan should have explicit process-exec DENY (writable dir)
.konan should have file-map-executable for Kotlin Native libs
```

Kotlin/Native downloads a prebuilt LLVM/clang toolchain into `~/.konan/dependencies/` and a native build links through it. That link step is an `execve` of a file under `~/.konan`, not a `dlopen`. `file-map-executable` — which is what the entry grants — does not cover `execve`; `process-exec` does, and it is explicitly denied.

If that is what happens, a Kotlin/Native compile inside cplt fails at link time on macOS, and the failure names a clang or lld path rather than the sandbox.

## Why this is its own issue

[...]

Issue #238, opened 2026-09-02
Deny write on the PATH-resolved bin and shim directories


Starefossen:
Several directories granted write+exec sit on the user's PATH ahead of `/usr/bin`, so an agent can leave a binary there that a later unsandboxed process picks up.

- mise AppDir (`sandbox_policy.rs:668-675`) covers `shims/` and `installs/<tool>/<ver>/bin`, both on PATH for mise users
- `PNPM_HOME` (`:1113-1117`) is the directory `pnpm setup` prepends to PATH
- `~/.deno` and `~/.bun` (`:863-874`) hold the `bin` directories their installers prepend
- `Library/pnpm` (`:967-971`), commented "global store + shims"

The profile emits allow-write and allow-exec for all of them with no compensating deny — the deny loop at `sandbox_profile.rs:809-822` only fires for `write && !exec`. `.cargo/bin` is deliberately read-only "to prevent a rogue agent from trojaning binaries", so the class is understood; it was not applied here.

A working exploit is six lines of shell: write `$HOME/.bun/bin/git`, `chmod 700`, exit normally, and the next process to resolve `git` through PATH runs it. #236 closes cplt's own parent-side spawns, but anything else the user runs — their own shell, another tool — still resolves through PATH.

## Proposal

Deny write on the bin/shim directories specifically, keeping the package/cache trees writable:

- `~/.bun/bin`, `~/.deno/bin`
- the mise `shims/` directory, and `installs/*/bin` if that can be expressed
- `PNPM_HOME` top level and `Library/pnpm`

This deliberately breaks global installs from inside the sandbox — `bun install -g`, `deno install`, `pnpm add -g`, `mise install <tool>`, `mise use -g`. That is the intent: an agent inside cplt should not be adding global tools to the host. Project-local dependency installs are unaffected — `npm install`, `cargo build`, `pip install` in a venv and `go build` write to the project or per-project caches, not to these directories.

Known cost to document: a repo whose `mise.toml` pins a toolchain the host does not have will no longer bootstrap inside the sandbox.

Verify per tool where installs actually land before writing the rules; mise in particular puts toolchains under `installs/` and shims under `shims/`, and only the latter is strictly PATH-resolved.

PR #330, opened 2026-09-04
fix(linux): re-bind both Copilot package dirs read-only

### LINKED ISSUE #328: Linux: neither Copilot pkg dir gets the write-deny SECURITY.md claims, and the SEA cache is writable+executable
macOS emits two write-denies for Copilot's package directories:

- `~/.copilot/pkg` — native modules (`keytar.node`, `pty.node`) — `src/sandbox_profile.rs:376`
- `~/Library/Caches/copilot/pkg` — the extracted SEA runtime — `src/sandbox_profile.rs:1067`

Linux has neither.

1. `~/.copilot` is granted `read+write+execute` wholesale (`src/sandbox_landlock.rs:766`) and Landlock cannot subtract `pkg` from an allowed tree, so the native modules are writable.
2. `~/.cache/copilot/pkg` is pushed as an `FsRule` with `write: false, execute: true` (`src/sandbox_landlock.rs:779`) — but the comment on that rule states it plainly: *"Write access is already inherited from the broader `~/.cache` allow"*. Landlock unions rules, so the effective posture is **write + execute** on the directory holding the runtime Copilot spawns.

That second one is the binary-drop pair (#238, #243, #321) in the one tree whose entire contents are code the host later executes.

SECURITY.md states the mitigation unconditionally in two places — the threat table (`| **Persistence via native modules** | Replace keytar.node with malware | Deny writes to ~/.copilot/pkg |`) and "Layer 0.5: Native module write protection". Both describe a protection Linux does not have. That is the defect the no-silent-grants rule exists for.

Fix: carry both paths in the bubblewrap `ro_protect` overlay, the same mechanism already backing the git- and agent-persistence denies, and state the bwrap dependency in SECURITY.md rather than leaving the unconditional claim standing.

Starefossen:
Closes #328.

## The gap

macOS emits two SBPL write-denies for Copilot's package directories: `~/.copilot/pkg` (native modules — `keytar.node`, `pty.node`) at `sandbox_profile.rs:376`, and `~/Library/Caches/copilot/pkg` (the extracted SEA runtime) at `sandbox_profile.rs:1067`.

Linux had neither.

1. `~/.copilot` is granted read+write+execute wholesale (`sandbox_landlock.rs:766`) so Copilot can reach its session store and native modules. Landlock cannot subtract `pkg` from an allowed tree.
2. `~/.cache/copilot/pkg` is pushed as an `FsRule` with `write: false, execute: true` — and the comment on that rule says the quiet part out loud: *"Write access is already inherited from the broader `~/.cache` allow"*. Landlock unions rules, so the effective posture was **write + execute** on the directory holding the runtime Copilot spawns.

Point 2 is the binary-drop pair that #238, #243 and #321 closed elsewhere today, in the one tree whose entire contents are code the host later executes. **I believe it is the last known instance of the class.** I looked for others while investigating: the remaining writable+executable trees are the deliberate opt-ins (`--allow-cache-exec`, `--allow-cache-exec-any`, `allow.exec`, the scratch dir), and `allow.exec` already refuses any tree overlapping a writable one. #323 (`~/.konan`) is the opposite polarity — exec denied where it is needed — and #324 is the execute bit on `~/.copilot`, which this PR does not touch.

## The fix

Both paths join the bubblewrap `ro_protect` overlay, the same mechanism already carrying the git-persistence denies, the agent config-dir denies and mise's `shims/`+`installs/`. Read-only rather than a deny mask: the runtime has to stay readable and executable. Nothing legitimate loses a write — SEA extraction runs before the sandbox is entered (`copilot_extract`, #166) and auto-update is disabled inside it (`--no-auto-update`).

The agent check lives inside `copilot_ro_protect_paths` rather than at the Linux-only call site in `sandbox.rs`, so the entire decision is one unit-testable function on both platforms instead of an `if` no macOS test can reach.

## SECURITY.md

Both claims were unconditional and both are now accurate. The threat table row and "Layer 0.5: Native module write protection" say what each platform actually enforces, name the second directory the old text omitted entirely, and state the two standing caveats of the bwrap overlay: without bubblewrap it is unenforced, and a path that does not exist at launch cannot be bound.

Where Landlock genuinely cannot express something, the document now says so rather than leaving a protection claimed that Linux does not have.

## Tests

- `copilot_ro_protect_paths_cover_both_package_dirs` — both paths in the set (runs on every platform).
- `copilot_ro_protect_paths_are_empty_for_other_agents` — no Copilot binds for Claude, Shell, OpenCode.
- `copilot_package_dirs_are_rebound_read_only_after_their_writable_parents` — the ordering assertion, which is the part that actually matters: bwrap mounts apply in argument order, so a read-only bind that lands before the writable parent buys nothing. Lives in the Linux-only `bubblewrap` module and runs in the Linux CI job's `cargo test --verbose`, not behind a skip.

**Mutation check:** dropped `home.join(".cache/copilot/pkg")` from the returned vector; `copilot_ro_protect_paths_cover_both_package_dirs` failed with `/Users/test/.cache/copilot/pkg must be in the bwrap read-only set`. Restored, green again.

---

## Merge order: #330 → #331 → #332

These three touch overlapping ground and must land in that order.

- **#330 first.** It extracts `ro_protect_paths()`, which #331's Linux coverage depends on, and it is the only one of the three with no behaviour trade-off.
- **#331 second.** It removes `assert!(Agent::Copilot.host_persistence_denies().is_empty())` from `host_persistence_denies_per_agent`; #332 keeps that line and adds its own Shell arm to the same test and the same `match`.
- **#332 last**, and it will need a hand resolution in `agent.rs` — both PRs edit the `host_persistence_denies` match arm and the same test — plus a CI re-run after.


Starefossen:
Follow-up commit, and the reason for it.

The mutation check ran on CI, not only locally, and it **passed** — deleting `ro_protect.extend(copilot_ro_protect_paths(...))` from the caller left Linux CI green. Both existing tests covered a helper: `copilot_ro_protect_paths_cover_both_package_dirs` asserts the function's return, and `copilot_package_dirs_are_rebound_read_only_after_their_writable_parents` passes that return straight into `build_bwrap_args`. Neither touched the assembly in `prepare_impl` that actually feeds the overlay, so the fix could have been silently unwired and no test would have said so.

`ro_protect_paths()` is now a function instead of an inline block, and `ro_protect_set_carries_the_copilot_package_dirs_for_copilot_only` asserts the assembled set — both dirs present for Copilot, absent for every other agent.

Re-ran the mutation with that test in place, on CI:

```
test sandbox::tests::ro_protect_set_carries_the_copilot_package_dirs_for_copilot_only ... FAILED
panicked at src/sandbox.rs:1120:
/home/test/.copilot/pkg must be in the bwrap read-only set, got
["/home/test/.local/share/mise/installs", "/home/test/.local/share/mise/shims",
 "/project/.agents/plugins", "/project/.cplt.toml", "/project/.git/hooks"]
test result: FAILED. 965 passed; 1 failed
```

Probe reverted, branch is back to the clean two commits, `linux-integration-tests` green — and the log confirms the Linux-only bwrap test executed rather than skipped:

```
test sandbox::bubblewrap::tests::copilot_package_dirs_are_rebound_read_only_after_their_writable_parents ... ok
```

Issue #222, opened 2026-08-27
Gradle plugin workers (ktlint, detekt) fail: JAVA_TOOL_OPTIONS does not reach WorkerExecutor forks, and SBPL's localhost token does not match ::ffff:127.0.0.1


Starefossen:
`ktlintCheck` and `detektMain` fail under the sandbox while `test` succeeds. A user diagnosed this precisely, and the diagnosis is correct — cplt's own source already documents the gap, just not anywhere a user would find it.

## The gap, as cplt describes it

`src/gradle_init.rs:18-21`:

> Scope: the script covers the daemon itself plus `Test`/`JavaExec` forks. `WorkerExecutor` process-isolation forks have no public configuration hook — those still require the plugin to propagate the environment (see JLLeitschuh/ktlint-gradle#1110).

The init script's own header names ktlint-gradle as the motivation, but the implementation only reaches `tasks.withType(JavaExec)` and `tasks.withType(Test)` (`src/gradle_init.rs:46-51`). Plugin workers using `WorkerExecutor` process isolation — ktlint-gradle, detekt — fork JVMs that never see `-Djava.net.preferIPv4Stack=true`, so their dual-stack sockets produce IPv4-mapped addresses.

The reporter confirmed this from the debug log: only the launcher logs `Picked up JAVA_TOOL_OPTIONS`; the worker daemon does not. They also ruled out port range, working directory, Gradle version (9.6.1 and 9.7.0), `--no-daemon`, `org.gradle.jvmargs`, and `java.net.preferIPv4Stack` in the JDK's `conf/net.properties` (not supported there).

## The second half: SBPL cannot match the mapped address with `localhost`

`src/sandbox_profile.rs:1457-1461` rests on the assumption that fails:

```rust
// Allow all localhost ports. With -Djava.net.preferIPv4Stack=true injected
// via JAVA_TOOL_OPTIONS (macOS), Java connections stay as pure IPv4 and
// "localhost:*" matches correctly. No need for "*:*" anymore.
sbpl!(sb, "(allow network-outbound (remote ip \"localhost:*\"))");
```

Tested against the kernel — a dual-stack listener, a client connecting to `::ffff:127.0.0.1`, under `sandbox-exec`:

| rule | result |
|---|---|
| `(remote ip "*:*")` | **OK** — the mapped address connects |
| `(remote ip "localhost:*")` | **FAIL** — `PermissionError` |

[...]

PR #336, opened 2026-09-04
feat(sandbox): route JVM traffic through the CONNECT proxy

Starefossen:
## What

The JVM has no `HTTP_PROXY` support. It reads `http.proxyHost` / `https.proxyHost`
system properties and nothing else. cplt injected neither, so every Gradle and Maven
dependency fetch went straight out through the kernel's `(allow network-outbound
(remote ip "*:443"))` allowance — never in the proxy log, never checked against the
blocklist. There was not one occurrence of `proxyHost`, `proxyPort` or `nonProxyHosts`
anywhere in `src/`.

This injects them into the existing `JAVA_TOOL_OPTIONS` block, using the proxy's
resolved port (OS-assigned by default, so it is read, not assumed):

```
-Dhttp.proxyHost=127.0.0.1 -Dhttp.proxyPort=<port> -Dhttp.nonProxyHosts=<list>
-Dhttps.proxyHost=127.0.0.1 -Dhttps.proxyPort=<port> -Dhttps.nonProxyHosts=<list>
```

The `JAVA_TOOL_OPTIONS` block also moves out of the `if let Some(scratch)` branch —
the proxy flags must be injected whenever the proxy runs, with or without a scratch
dir. Behaviour with no proxy is byte-identical to before.

`https.nonProxyHosts` is not read by the JDK's own `DefaultProxySelector`, but Gradle
documents and honours it. It costs one flag, so both are set.

## nonProxyHosts

`localhost|*.localhost|127.*|[::1]|::1|0.0.0.0`

This is deliberately the same set `proxy.rs::handle_connect` recognises as loopback —
`localhost`, `*.localhost` (RFC 6761), any IP literal that `is_loopback()` — plus
`0.0.0.0`, the wildcard bind address a JVM dev server or test container reports for
itself. Both bracketed and bare `::1` because Java's URI host keeps the brackets on
some JDK paths and drops them on others.

Justification for that boundary rather than a longer list: loopback is only reachable
*through* the proxy under `--allow-localhost`, so bypassing it changes nothing the
proxy would have permitted, and it must bypass or the Gradle daemon connection, the
Kotlin daemon, every local dev server and every Testcontainers-mapped port break.
Anything wider (`*.local`, `169.254.*`, RFC1918 ranges) would silently exempt hosts
from logging and filtering, which is the opposite of the point of this PR. Those cases
belong in `proxy.allow_private_domains`, which already exists — see the breakage
section.

## Gradle daemon and worker forks — what is and is not proxied

Checked, not assumed:

| JVM | Proxied? | How |
| --- | --- | --- |
| Gradle launcher | yes | `JAVA_TOOL_OPTIONS` |
| Gradle daemon (started inside cplt) | yes | daemon inherits the launcher's environment; verified in the run below, where dependency resolution happens in the daemon |
| Gradle daemon started *outside* cplt and reused | **no** | `JAVA_TOOL_OPTIONS` is not part of Gradle's daemon-compatibility check, so a pre-existing daemon can be reused without the properties. Unchanged from today's behaviour (it is not proxied now either), but worth knowing: `--no-daemon`, or a daemon first started inside cplt, is proxied. |
| `Test` / `JavaExec` forks | yes, with `sandbox.gradle_init = true` | the init script now forwards the properties from the daemon's system properties into `jvmArgs` |
| `WorkerExecutor` process-isolation forks | **no** | this is #222. `JAVA_TOOL_OPTIONS` does not reach them and `WorkerExecutor` has no public fork-configuration hook, so the init script cannot reach them either. Not addressed here. A worker fork that resolves dependencies or calls out still goes direct on `*:443` and still does not appear in the proxy log. |

The init script reads the properties rather than baking in a port: the port is
OS-assigned and `SCRIPT_BODY` is static content that persists in
`~/.gradle/init.d/`. Baking a port in would leave a stale one for the next run. It
stays inert when no proxy is running.

## What a normal Gradle/Maven build contacts, and whether it passes

Probed with `cplt --agent copilot check net --no-connect`:

| Host | Why | Blocklist-only (default) | `--default-allowlist` |
| --- | --- | --- | --- |
| `repo.maven.apache.org` | Maven Central | ALLOWED | ALLOWED |
| `repo1.maven.org` | Central's other name, used by many declarations | ALLOWED | **BLOCKED-ALLOWLIST** |
| `plugins.gradle.org` | plugin portal | ALLOWED | ALLOWED |
| `services.gradle.org` | wrapper distribution download, version check | ALLOWED | **BLOCKED-ALLOWLIST** |
| `dl.google.com` | Android/Google Maven | ALLOWED | **BLOCKED-ALLOWLIST** |
| `api.foojay.io` | Gradle toolchain (JDK) provisioning | ALLOWED | **BLOCKED-ALLOWLIST** |
| `scans-in.gradle.com` | build scans / Develocity | ALLOWED | **BLOCKED-ALLOWLIST** |
| `maven.pkg.github.com` | GitHub Packages | ALLOWED | ALLOWED |

Nothing in `blocked-domains.txt` is on a JVM build path — the list is webhook,
paste, tunnel, IP-echo and telemetry endpoints. **Under the blocklist-only default,
which is what ships, no host a normal Gradle or Maven build contacts is blocked.**
`services.gradle.org` missing from `PACKAGE_REGISTRY_DOMAINS` therefore does not
matter today; it matters the moment anyone turns on `--default-allowlist`, and so do
the four other rows. Left alone here — extending the allowlist is a separate decision
with its own blast radius, and this PR is already a behaviour change.

Port: `allowed_ports` defaults to `[443]` and every host above is HTTPS. A
`repositories { maven { url "http://..." } }` declaration gets BLOCKED-PORT, but the
kernel only ever allowed `*:443` anyway, so that is not a regression.

No TLS interception is involved — `handle_connect` is a plain CONNECT tunnel — so
there is no JVM truststore problem.

## What this breaks, and the remedy for each case

**Internal Maven repositories on private IPs.** An in-house Nexus or Artifactory, or a
self-hosted Develocity, that resolves to RFC1918/CGNAT space is reached *directly*
today (the kernel allows `*:443` to anywhere). Sent through the proxy it hits the SSRF
guard. This is the same guard curl, npm and pip have hit all along, because they have
had `HTTP_PROXY` set the whole time; JVM users were the anomaly, not the casualty.

`allow_private_domains` matches on domain, so the two ways of configuring a repository
do **not** have the same answer. Determined with `cplt check net`, not reasoned about:

**Repository configured by DNS name — allowable.**

```
$ cplt --allow-private-domain nip.io check net --no-connect nexus.10.20.30.40.nip.io
nexus.10.20.30.40.nip.io:443: ALLOWED
```

```toml
[proxy]
allow_private_domains = ["intern.example.com"]   # suffix match, covers subdomains
```

**Repository configured as a bare IP literal — NOT allowable, by any key.**

```
$ cplt --allow-private-domain 10.20.30.40 check net --no-connect 10.20.30.40
10.20.30.40:443: BLOCKED
```

Listing the literal does nothing. The private check for an IP literal is the *pre-DNS*
gate in `classify_connect`, which never consults the private-domain list — that list is
read only by the post-DNS guard in `classify_resolved`. There is no flag that reaches
it. `https://10.20.30.40/repository/maven-public/` in a `build.gradle` or
`settings.xml`, which is not rare for internal infrastructure, needs a DNS name; there
is no other route. Pinned by
`ip_literal_stays_private_blocked_even_when_listed_as_a_private_domain` so the
documented claim cannot rot.

Saying that plainly is the point. A remedy that silently works for only half the cases
is the no-silent-grants defect wearing a new hat.

**The refusal text named a flag that does not exist.** It said
`--allow-private-domains <DOMAIN>`; the flag is singular, and the plural form exits
with `error: unexpected argument`. Fixed in both the `check net` verdict and the
post-DNS explanation. Both now also name the config key, and the IP-literal verdict
says the case has no remedy rather than sending the user to a flag that cannot help.

**The 403 the build actually sees now names the fix**, which is worth more than a
paragraph someone has to find:

```
403 Forbidden — Private target blocked by cplt. For a trusted internal host, add its
DNS name to proxy.allow_private_domains (or pass --allow-private-domain). An
IP-literal target cannot be allowed: give the host a name.
```

**Documented where a Gradle user will hit it**, not in SECURITY.md: a new
`docs/known-impacts.md` section ("Internal Maven/Gradle repositories on private IPs")
placed next to *Private registries*, a row and a paragraph in the README's Known
impacts, and the existing *Gradle plugin workers* section extended to cover the proxy.

This is flagged rather than worked around: silently widening `nonProxyHosts` to let
private hosts skip the proxy would restore the build and destroy the property the PR
exists for.

## The `proxy.forced` precondition

`sandbox_landlock.rs` carries the comment "Gradle is already unusable under
`proxy.forced`", because forced mode drops the `*:443` allowance and the JVM had no
other path. With this change the JVM has one. That does not make `proxy.forced`
turnkey for JVM users — `WorkerExecutor` forks (#222) still have no route, and
internal repositories need `allow_private_domains` — but it removes the reason it was
impossible rather than merely awkward.

`README.md:812` claims the macOS proxy log is complete under `proxy.forced`. That
claim is **left alone**. It was true only because JVM traffic was denied rather than
routed, and it is still not true for a better reason: `WorkerExecutor` forks are
unproxied, so under a non-forced run their traffic is absent from the log, and under a
forced run they fail rather than appear. Revisit when #222 has an answer.

## Verification

**Real Gradle build, real proxy log.** `gradle --no-daemon deps` resolving Guava from
Maven Central, inside cplt with `--with-proxy`:

```
Picked up JAVA_TOOL_OPTIONS: ... -Dhttp.proxyHost=127.0.0.1 -Dhttp.proxyPort=61548
  -Dhttp.nonProxyHosts=localhost|*.localhost|127.*|[::1]|::1|0.0.0.0
  -Dhttps.proxyHost=127.0.0.1 -Dhttps.proxyPort=61548 -Dhttps.nonProxyHosts=...
BUILD SUCCESSFUL in 6s

$ cat proxy.log
2026-09-04T19:21:38Z CONNECT repo.maven.apache.org:443 CONNECTED
... (7 CONNECTs, one per artifact)
```

**Baseline on the same machine, same command, `main`'s binary, cache wiped first:**
build succeeds, all 7 jars land in the Gradle cache, and the proxy log is **empty**.
That is the gap this PR closes, measured rather than argued.

(The interleaved timestamps in the log are a pre-existing concurrent-write artifact in
the log writer, not from this change.)

**Tests.** `env_proxy_running_injects_jvm_proxy_properties` asserts the six flags and
the loopback bypass are present with a proxy port;
`env_no_proxy_omits_jvm_proxy_properties` asserts `proxyHost`/`proxyPort`/
`nonProxyHosts` are absent without one;
`env_proxy_properties_injected_without_scratch_dir` pins that the flags do not depend
on a scratch dir; `env_pass_env_java_tool_options_still_suppresses_proxy_injection`
keeps the `--pass-env` escape hatch working.
`gradle_init::script_forwards_proxy_properties_to_forks` pins the fork forwarding and
that no port is baked in.

**Mutation check.** Disabling the injection (`if let Some(port) = proxy_port && false`)
turns `env_proxy_running_injects_jvm_proxy_properties` and
`env_proxy_properties_injected_without_scratch_dir` red — `test result: FAILED. 325
passed; 2 failed` — while `env_no_proxy_omits_jvm_proxy_properties` stays green, which
is what it should do. Restored, all 327 pass.

**Gates.** `cargo fmt --check`, `cargo test --lib` (923), `cargo test --test
unit_tests` (327), `cargo clippy --all-targets -- -D warnings` clean, plus
integration/e2e/e2e_projects/e2e_guards/e2e_git_hardening (397) all green.


manderson23:
One comment about Maven is that, from my experience testing in the past, you also need to add `-Daether.connector.http.useSystemProperties=true` for it to make use of the `http.proxyHost / https.proxyHost` system properties. 

Though this is not recommended for using Maven with proxies. Explicitly updating config files is preferred which isn't ideal with cplt. Maven 4 changes will at least mean environment variables can be referenced for proxy in settings.xml.

Issue #298, opened 2026-09-04
cplt doctor hangs forever when an agent binary on PATH does not exit


Starefossen:
Found while auditing test isolation for #245, but this is a product bug, not a test bug.

`discover_agents` (`src/discover.rs:207-238`) probes `copilot`, `opencode`, `antigravity`/`agy` and `claude` by running each with `--version` through `.output()`. There is no timeout on any of them. If a single one of those binaries is on `PATH` and does not exit, `cplt doctor` never returns.

## How it shows up

`sample(1)` on a wedged child:

```
cplt::run_doctor (main.rs:4297) -> discover::discover_all -> discover_agents
  -> std::process::Command::output -> poll
```

The trigger here was `claude --version`, which does not return when invoked from inside a Claude Code session. Two wedged `cplt doctor` processes from unrelated sessions had been running for nearly two hours when this was found. In the test suite it presents as `cargo test --test e2e` hanging rather than failing — 683 seconds and still going, against 5 seconds once the cause is removed, because `e2e.rs` calls doctor six times.

The reason this is easy to miss: the probe only wedges when a *particular* binary on *your* PATH misbehaves. CI has no agent CLIs installed, so it never sees it.

## Why it deserves a real fix rather than a test workaround

Any user with a hung or misbehaving agent binary on `PATH` gets a `cplt doctor` that hangs with no output and no indication which probe is stuck. Doctor is the command people run when something is already wrong, so it is exactly the wrong place to hang.

## Shape of the fix

Wait-with-timeout around each probe, with the timed-out agent reported as undetected or as an explicit "probe timed out" row rather than silently omitted. A per-probe timeout is better than one overall, so a single bad binary does not hide the others.

Worth deciding as part of it: whether a timed-out probe should be visible in doctor output. Reporting it is more useful than treating it as absent, since "agent not detected" and "agent did not answer" are different problems with different fixes.

This is a behaviour change in a security tool, so it wants its own PR rather than riding along with unrelated work.

PR #345, opened 2026-09-04
fix(sandbox): grant the developer dir xcode-select selects

### LINKED ISSUE #342: Sandbox grants a hardcoded Xcode path, so xcrun shims break on any non-default Xcode
`TOOL_READ_DIRS` (`src/sandbox_policy.rs:331`) grants a literal path:

```rust
// Xcode.app developer tools — needed when xcode-select points to the full
// Xcode install instead of standalone CommandLineTools. /usr/bin/git and
// ...
"/Applications/Xcode.app/Contents/Developer",
```

The comment says exactly why it exists. The path is hardcoded, and `xcode-select` does not have to point there.

## Impact

On macOS, `/usr/bin/git`, `/usr/bin/clang`, `/usr/bin/make`, `/usr/bin/python3` and the rest are xcrun shims. Each dlopens `libxcrun.dylib` from whatever `xcode-select` points at. If that is anything other than `/Applications/Xcode.app` — a versioned install, a beta, an Xcode in a non-default location — the sandbox does not grant it and every one of those tools dies inside the sandbox with:

```
xcrun: error: unable to load libxcrun (dlopen(/Applications/Xcode_26.6.app/Contents/Developer/usr/lib/libxcrun.dylib, 0x0005):
  tried: '...libxcrun.dylib' (file system sandbox blocked open()), ...)
```

This is **independent of the guards**. Anyone whose only git is `/usr/bin/git` and whose `xcode-select` is a versioned Xcode cannot run git inside cplt at all, guard on or off. Users with the Command Line Tools, or with a PATH git from Homebrew/mise/asdf/nix, never see it — which is presumably why it has gone unnoticed.

Observed on the project's own macOS CI, where `xcode-select` points at `/Applications/Xcode_26.6.app`.

## Fix

Resolve the path instead of hardcoding it: `/var/db/xcode_select_link` is a symlink to the selected developer directory, so `std::fs::read_link` gives the real answer. Pure Rust, no subprocess, no `xcode-select` spawn. Keep the literal as a fallback for the case where the symlink is absent (Command Line Tools installs, where `/Library/Developer/CommandLineTools` is already granted).

## Not this

#338 stops the **git guard** from making this worse: the wrapper was overriding the agent's working PATH git with `/usr/bin/git`, so the shim broke git for people who had a perfectly good one. That is a separate bug with a separate fix, already merged. It does not repair the shim, and a machine with no git but `/usr/bin/git` is still broken after it.

Found while diagnosing #338.

Starefossen:
Fixes #342.

`TOOL_READ_DIRS` granted the literal `/Applications/Xcode.app/Contents/Developer`. On macOS `/usr/bin/git`, `/usr/bin/clang`, `/usr/bin/make` and `/usr/bin/python3` are xcrun shims that `dlopen` `libxcrun.dylib` from whatever `xcode-select` points at, so on a versioned, beta or relocated Xcode the sandbox denies that open and each shim exits 1 before the real tool runs. A user whose only git is `/usr/bin/git` has no git inside the sandbox at all. Independent of the git guard fixed in #338.

## The fix

`xcode_developer_dir_from()` reads `/var/db/xcode_select_link` with `std::fs::read_link`. No subprocess: `xcode-select -p` would put an unbounded child on the profile-generation path in the parent, which is what made `cplt doctor` hang (#298).

- The literal stays in `TOOL_READ_DIRS` as the fallback when the link is absent.
- A selection already covered by `TOOL_READ_DIRS` emits no second rule.
- The resolved target goes through `tool_override_path_is_safe` and `validate_sbpl_path`, so a link at `/`, `$HOME` or an ancestor of it is dropped rather than granted. The link is root-owned, but a grant that follows a symlink gets the same widening check every other resolved path in this codebase gets.

## What the file actually is

Checked on the dev machine before writing the code: `/var/db/xcode_select_link` is a **symlink** (`lrwxr-xr-x root:wheel`), not a plain file containing a path, and it points at `/Library/Developer/CommandLineTools`. `read_link` is therefore the right call. A non-symlink would make `read_link` return `Err`, which falls back to the literal — degraded, not broken.

## Command Line Tools

No extra grant needed. When `xcode-select` points at `/Library/Developer/CommandLineTools`, that path is already in `TOOL_READ_DIRS`; the dedupe check skips it so the profile gains no duplicate rule. Covered by `command_line_tools_selection_emits_no_duplicate_grant`.

## Configurations

Verified by test, on any machine, by pointing `emit_tool_dirs` at a symlink under test control:

- versioned Xcode (`/Applications/Xcode_26.6.app/Contents/Developer`) — granted read + `file-map-executable`
- link absent — falls back to the literal
- Command Line Tools — no duplicate grant
- link to `/`, `$HOME`, `/Users`, `/tmp` — not granted

Verified on real hardware: only the Command Line Tools configuration, which is what this machine has. The versioned-Xcode case is the one CI reproduces; the profile assertion above is the same rule the failing CI run needed.

Reasoned about, not executed: whether a real versioned Xcode's shims recover once the grant lands. The dlopen path in the CI error sits directly under the granted subpath, so the grant covers it.

## Mutation check

Replaced the resolution with `Option::<PathBuf>::None` (the literal-only behaviour). `sandbox::profile::tests::versioned_xcode_developer_dir_is_granted` FAILED, 934 passed. Restored; 935 pass.

## Gates

`cargo fmt`, `cargo test --lib` (935), `cargo test --test unit_tests` (329), `cargo clippy --all-targets -- -D warnings` (clean), and `cargo test --test integration` (65) — all green on macOS.

### DIFF
--- a/src/sandbox_policy.rs
+++ b/src/sandbox_policy.rs
@@ -354,17 +354,27 @@ pub const XCODE_SELECT_LINK: &str = "/var/db/xcode_select_link";
 /// them on any machine with a versioned or relocated Xcode — a user whose only
 /// git is `/usr/bin/git` then has no git inside the sandbox at all (#342).
 ///
-/// Returns `None` when the link is absent, when it is relative, when it names a
-/// path that would widen the sandbox ([`tool_override_path_is_safe`] — the link
-/// is root-owned, but a grant that follows a symlink still gets the same sanity
-/// check every other resolved path in this codebase gets), or when the target
-/// cannot be safely interpolated into SBPL.
+/// Returns `None` when the link is absent, when it is not a symlink, when its
+/// target does not exist, when it names a path that would widen the sandbox
+/// ([`tool_override_path_is_safe`] — the link is root-owned, but a grant that
+/// follows a symlink still gets the same sanity check every other resolved path
+/// in this codebase gets), or when the target cannot be safely interpolated
+/// into SBPL.
+///
+/// The target is canonicalized before it is checked and before it is granted.
+/// `tool_override_path_is_safe` compares paths textually and documents that its
+/// argument is already canonical, so a link target carrying `..` — or one whose
+/// own parents are symlinks — would otherwise slip past it and be emitted as a
+/// grant on whatever it really resolves to.
 ///
 /// The caller is responsible for skipping a target already covered by
 /// [`TOOL_READ_DIRS`] — the Command Line Tools case, where the selected
 /// directory is `/Library/Developer/CommandLineTools` and is already granted.
 pub fn xcode_developer_dir_from(link: &Path, home: &Path) -> Option<PathBuf> {
-    let target = std::fs::read_link(link).ok()?;
+    // `read_link` rather than `canonicalize` alone: it fails on a regular file,
+    // so a `/var/db/xcode_select_link` that is not a symlink falls back to the
+    // literal instead of granting a subpath on the file itself.
+    let target = std::fs::canonicalize(std::fs::read_link(link).ok()?).ok()?;
     if !target.is_absolute()
         || !tool_override_path_is_safe(&target, home)
         || validate_sbpl_path(&target).is_err()
--- a/src/sandbox_policy.rs
+++ b/src/sandbox_policy.rs
@@ -328,10 +328,51 @@ pub(super) const TOOL_READ_DIRS: &[&str] = &[
     "/Library/Developer/CommandLineTools",
     // Xcode.app developer tools — needed when xcode-select points to the full
     // Xcode install instead of standalone CommandLineTools. /usr/bin/git and
-    // other shims load libxcrun.dylib from this path.
+    // other shims load libxcrun.dylib from this path. Only the *default*
+    // install location; the directory actually selected is resolved at profile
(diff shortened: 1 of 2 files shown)


Starefossen:
Addressed the canonicalization finding in 
`fix(sandbox): canonicalize the resolved developer dir before granting it`.

`tool_override_path_is_safe` compares textually and documents that its argument is already canonical, so a link target carrying `..` (or one whose parents are symlinks) would have slipped past it. The target is now canonicalized before the check and before it is emitted. `read_link` stays in front of `canonicalize` so a `xcode_select_link` that is a regular file falls back to the literal rather than granting a subpath on the file itself — covered by `regular_file_xcode_select_link_grants_nothing_extra`.

Mutation check on the new behaviour: dropping the `canonicalize` call makes `xcode_select_link_to_an_unsafe_root_is_not_granted` FAIL (935 passed, 1 failed); restored, 936 pass. The original mutation still holds too — reverting the emit to literal-only fails `versioned_xcode_developer_dir_is_granted`.

Note the tests now build real directories under a temp $HOME, since a canonicalizing resolver will not grant a path that does not exist.

PR #352, opened 2026-09-04
fix(doctor): grade Windows-side tools under WSL instead of ticking them green

### LINKED ISSUE #271: doctor: gi feil, ikke grønt lys, når verktøy løses til Windows-stier under WSL
## Problem

En bruker på Windows/WSL2 fikk `copilot --version` til å feile med:

```
GitHub Copilot CLI: no platform package found. Reinstall with `npm install -g @github/copilot` to fetch the package for your platform.
```

`cplt doctor` hadde all informasjonen som trengtes for å diagnostisere dette, men graderte det som grønt:

```
[doctor] Agents
  ✓ Copilot (copilot): /mnt/c/Users/N129069/AppData/Roaming/npm/copilot

[doctor] Tools
  ✓ yarn: /mnt/c/Users/N129069/AppData/Roaming/npm/yarn
  ✓ gradle: /mnt/c/apps/Gradle/gradle-8.10.2/bin/gradle
...
[doctor] All critical checks passed ✓
```

Alle disse stiene peker på Windows-binærer, arvet inn i WSL via `interop.appendWindowsPath`. De kan ikke kjøre i Linux-namespacet, og de kan ikke sandboxes av Landlock. `All critical checks passed` er feil konklusjon.

## Rotårsak hos brukeren

`apt install nodejs` ga `/usr/bin/node`, men Ubuntu-pakken bunter ikke `npm`. `npm` løste derfor til Windows-installasjonen, `npm root -g` ga `C:\Users\...\AppData\Roaming\npm\node_modules`, og `npm install -g @github/copilot` hentet win32-x64-plattformpakken. Feilmeldingen fra Copilot CLI var teknisk korrekt, men ledet brukeren til å reinstallere om og om igjen med samme resultat.

Delvis skygging er det farlige tilfellet: Windows-PATH legges bakerst, så en Linux-variant vinner alltid når den finnes. Mangler den for ett enkelt verktøy, blir akkurat det verktøyet stille hentet fra Windows.

## Forslag

Når `uname -r` inneholder `microsoft` (WSL), bør doctor behandle enhver løst verktøysti som starter med `/mnt/` som feil, ikke som OK:

```
[doctor] Agents
  ✗ Copilot (copilot): /mnt/c/Users/.../npm/copilot — Windows-binær, kan ikke kjøre i WSL
      Fiks: curl -fsSL https://gh.io/copilot-install | bash
```

Gjelder minst `copilot`, `node`, `npm`, `gh` og `git`. For byggverktøy (`gradle`, `yarn`, `java`) holder det trolig med advarsel og en note om at en build startet derfra kjører som Windows-prosess, utenfor sandkassen.

Sjekken finnes allerede — den printer stien. Det som mangler er graderingen.

Starefossen:
Closes #271 — the tools half of the WSL grading bug.

## What was wrong

The agents half was already fixed in #204: `is_wsl_interop_binary` grades a Windows-side agent red, so the issue's headline example no longer reproduces. The tools loop never got the same treatment — it printed an unconditional green tick for every tool `which` resolved, so a WSL user whose tools came from the Windows side through interop got a full green report and `All critical checks passed` for binaries that cannot run in the Linux namespace and cannot be sandboxed by Landlock.

Worse, `npm` was not in `TOOLS_TO_CHECK` at all. That is the root cause the reporter actually hit: Ubuntu's `nodejs` package does not bundle npm, so `node` resolved Linux-side and looked fine while `npm` alone fell through to the Windows install, made `npm root -g` a Windows path, and pulled the win32 platform package for the Copilot CLI. Partial shadowing is the dangerous case, and doctor could not even see it.

## What changed

- `npm` added to `TOOLS_TO_CHECK`.
- The tools loop applies `is_wsl_interop_binary`, the same predicate and the same WSL gate the agents loop uses. `gh`, `git`, `node` and `npm` are red and fail the critical checks; everything else (gradle, yarn, java, …) gets a warning saying a build started from it runs as a Windows process outside the sandbox. That split is the one the issue proposes.
- The per-tool lines moved into `tool_lines(&[ToolInfo], wsl) -> (Vec<String>, bool)` so the behaviour is testable. `print_report` prints exactly what it returns and cannot drift from it.

Off WSL, `/mnt/c` stays an ordinary mount and a tool there is still green — the gate is `is_wsl()`, unchanged.

## Tests

Four in `discover::wsl_tool_report_tests`, asserting on the emitted line rather than the predicate:

- `windows_side_critical_tool_is_reported_red_and_fails_critical` — a `/mnt/c/Users/.../npm/npm` produces a ✗, names the path, and returns `critical_ok == false`.
- `windows_side_build_tool_is_warned_not_failed` — a Windows gradle produces ⚠ and leaves the critical checks passing.
- `mnt_path_outside_wsl_stays_green` — same path with `wsl == false` stays ✓.
- `npm_is_checked`.

**Mutation check:** replacing the predicate with `if true` (restoring the unconditional green) fails `windows_side_critical_tool_is_reported_red_and_fails_critical` and `windows_side_build_tool_is_warned_not_failed`, the latter with `expected a warning sign, got:   ✓ gradle: /mnt/c/apps/Gradle/gradle-8.10.2/bin/gradle`. Restored, all four pass.

## Gates

`cargo fmt --check`, `cargo test --lib` (959), `cargo test --test unit_tests` (331), `cargo clippy --all-targets -- -D warnings` — all clean.

## Not done

The `/etc/wsl.conf` `automount.root` ceiling documented on `is_windows_interop_path` still applies: a relocated mount root is not detected. Unchanged by this PR, and the failure mode stays the preferred one — the tool fails later with its path in the message rather than being refused while working.

### DIFF
--- a/src/discover.rs
+++ b/src/discover.rs
@@ -322,9 +322,9 @@ const TOOLS_TO_CHECK: &[&str] = &[
 /// These are the tools cplt itself and the agents depend on: a Windows `npm`
 /// installs the win32 platform package into the Windows tree, which is exactly
 /// how a user ended up with a Copilot CLI that could not start (#271). The
-/// build tools (`gradle`, `yarn`, `java`, ...) are only warned about: a build
-/// started from one runs as a Windows process outside the sandbox, which is
-/// worth saying but is not a broken install.
+/// rest — build tools, language runtimes, the `app_dirs()` applications — are
+/// only warned about: running one starts a Windows process outside the sandbox,
+/// which is worth saying but is not a broken install.
 const WSL_CRITICAL_TOOLS: &[&str] = &["gh", "git", "node", "npm"];
 
 use crate::sandbox::app_dirs;
@@ -485,7 +485,7 @@ fn tool_lines(tools: &[ToolInfo], wsl: bool) -> (Vec<String>, bool) {
             } else {
                 format!(
                     "  {}\u{26a0}{} {}: {} is a Windows install reached through WSL interop; \
-                     a build started from it runs as a Windows process, outside the sandbox.",
+                     running it starts a Windows process, outside the sandbox.",
                     ui::stdout_color(ui::YELLOW),
                     ui::stdout_color(ui::RESET),
                     tool.name,
@@ -2193,7 +2193,7 @@ mod wsl_tool_report_tests {
         );
     }
 
-    /// A build tool is only warned about — it is not a broken install.
+    /// A non-critical tool is only warned about — it is not a broken install.
     #[test]
     fn windows_side_build_tool_is_warned_not_failed() {
         let tools = vec![tool(
--- a/src/discover.rs
+++ b/src/discover.rs
@@ -313,8 +313,19 @@ pub fn discover_agents() -> Vec<AgentInfo> {
 // ── Tool discovery ──────────────────────────────────────────────
 
 const TOOLS_TO_CHECK: &[&str] = &[
-    "gh", "git", "node", "cargo", "python3", "java", "go", "gradle", "yarn",
+    "gh", "git", "node", "npm", "cargo", "python3", "java", "go", "gradle", "yarn",
 ];
+
+/// Tools whose Windows-side resolution under WSL is a hard failure, not a
+/// warning.
+///
+/// These are the tools cplt itself and the agents depend on: a Windows `npm`
+/// installs the win32 platform package into the Windows tree, which is exactly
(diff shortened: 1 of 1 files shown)


Starefossen:
Good catch — the warning now reads `running it starts a Windows process, outside the sandbox`, which is true of `python3` and the `app_dirs()` applications as well as a build tool. The const comment and the test name are reworded to match.

Issue #144, opened 2026-07-07
Design: subscribable domain lists — curated defaults + org/team subscriptions (extends #10)


Starefossen:
# PRD: Subscribable domain lists — curated defaults + org/team subscriptions

> Design PRD. Grounded in the current code. **Verdict: worthwhile** — but with a hard bifurcation between blocklist and allowlist subscriptions, because they are different *risk classes*. Subsumes and extends #10 (auto-updatable blocklist).

## Verdict up front

**Yes, build it — in two tiers, because "like adblock lists" is only half right:**

- **Blocklist subscriptions** (deny known-bad, like uBlock/EasyList): low-risk, tighten-only, fail-open acceptable. **Clear win, do first.** This is #10.
- **Allowlist subscriptions** (define what's permitted): high value for org governance (NAV publishes one canonical approved-domains list, all teams subscribe), but **this is cplt's security boundary** — a compromised or MITM'd allowlist opens an exfiltration channel for every subscriber. Must be security-engineered (signed, fail-to-cache, global-only). **Do carefully, second.**

The adblock analogy is the right UX inspiration but has a critical asymmetry: **adblock lists are blocklists that fail *open* (worst case: an ad shows).** A cplt allowlist that fails open or is tampered = data exfiltration. So the trust/fail model must be *inverted* for the allowlist case.

## Why it's worthwhile

- **Org governance (the NAV case):** today each team hand-maintains `allowed_domains`. A subscription lets NAV maintain ONE canonical list (internal APIs, approved registries, agent infra) that every team subscribes to — consistent, centrally updated, no per-team drift. This is exactly NAV's managed-policy model and the natural completion of the secure-by-default epic (#71).
- **Fresh threat coverage (#10):** the exfil blocklist (96 curated entries from LOTS/Socket/Mend) goes stale; the LotAI C2 pattern moves fast. A maintained, subscribable blocklist keeps it current without a cplt release.
- **Ecosystem lists:** shareable per-stack allowlists (node-dev, python-dev, per-agent) — the community-list dimension.
- **Decoupling:** domain policy changes far more often than cplt; subscriptions update it independently of releases.

## What already exists (build on, don't reinvent)

- **List loading + hot-reload** (`src/proxy.rs`): `allowed_domains_file`/`blocked_domains` load from files, are TTL-reloaded, and the effective allowlist is already **built-in (#52) ∪ reloadable file**. A subscription is "keep a local cache file fresh from a URL"; the existing loader then consumes it unchanged.
[...]

Issue #62, opened 2026-05-14
feat: MCP server sandboxing (isolate tool connectors)


Starefossen:
## Value Proposition

MCP (Model Context Protocol) servers are the #1 attack surface for agentic IDEs in 2025. Cursor's CurXecute and MCPoison CVEs (CVE-2025-54135/54136) demonstrated that unsandboxed MCP servers lead to RCE. cplt should offer MCP server isolation as a first-class feature.

## Context

- Copilot CLI supports MCP tools (via `copilot-setup-steps.yml` and runtime config)
- MCP servers run as child processes with full host access by default
- Cursor's vulnerabilities: forged workspace files loaded attacker-controlled MCP servers
- VS Code now requires explicit trust per MCP server

## Proposal

When cplt detects MCP server processes spawned by the agent:

1. **Inherit sandbox restrictions** — MCP servers should not escape the agent's sandbox
2. **Additional restrictions** — MCP servers could be further restricted (no project write, network-only to specific endpoints)
3. **Visibility** — log which MCP servers are active and what they access

### Implementation approaches

| Approach | Complexity | Effectiveness |
|----------|-----------|---------------|
| Sandbox inherits to children (current) | Already done ✅ | Good — children inherit SBPL/Landlock |
| Explicit MCP server policy in .cplt.toml | Medium | Better — per-server restrictions |
| MCP traffic routing through proxy | High | Best — full visibility and control |

## Note

Since cplt sandboxes the entire process tree (not just the agent binary), MCP servers spawned as children already inherit restrictions. The value-add here is:
- Making this explicit and documented
- Adding per-server policy configuration
- Providing visibility into MCP server activity

PR #353, opened 2026-09-04
fix(proxy): add the Claude hosts the built-in allowlist was missing

### LINKED ISSUE #218: Claude's built-in domain allowlist is missing platform.claude.com, claude.com and mcp-proxy.anthropic.com
The built-in Claude allowlist (`src/agent.rs:96-99`) contains four domains:

```rust
"api.anthropic.com",
"console.anthropic.com",
"claude.ai",
"statsig.anthropic.com",
```

Claude Code now also reaches `platform.claude.com`, `claude.com` and `mcp-proxy.anthropic.com`. None is on the list, so with `default_allowlist` on — the fail-closed configuration we recommend — those connections are blocked.

## Why this is easy to misdiagnose

The failure does not look like a domain block. A user hits it while the agent is doing something else entirely, sees a generic API error, and the proxy log shows `BLOCKED-ALLOWLIST` only if they thought to enable `--proxy-log-level all`. It was very nearly attributed to the tunnel-timeout bug fixed in #194, because the symptoms overlap.

## What to check before adding

Do not just append the three names. Establish which hosts Claude Code actually uses today and why, so the list is defensible rather than accumulated:

- Run a Claude Code session under `--proxy-log-level all` with `default_allowlist` on and collect every `BLOCKED-ALLOWLIST` line.
- Distinguish what is required for the agent to function from what is telemetry or optional. `statsig.anthropic.com` is already on the list, so the project has evidently accepted some telemetry — but that should be a deliberate line, and the blocklist in `blocked-domains.txt` exists partly to strip telemetry.
- `mcp-proxy.anthropic.com` is MCP-related and interacts with #62 (MCP server sandboxing). Adding it silently widens what an MCP connector can reach.

## Worth fixing at the same time

The other agents' built-in lists have the same staleness risk and nobody has audited them recently — Copilot, OpenCode, Gemini, Antigravity, Pi. A per-agent list that drifts is a fail-closed policy that quietly becomes a support burden.

Consider whether these lists should carry a "last verified" date, or whether #144 (subscribable domain lists) is the real answer for keeping them current.

Reported by a user hitting API errors under the default policy.

Starefossen:
Closes #218.

Note on wording: this repo's publishing guard rejects any text containing the literal Anthropic domain string, so hosts under it are named by role below (`mcp-proxy…`, `statsig…`, the main API host). The exact strings are in the diff.

## Provenance, since that was the point of the issue

The issue asked not to just append three names. Two independent sources, both checked on 2026-09-04:

1. Anthropic's published **Network access requirements** table (`code.claude.com/docs/en/network-config`), which lists every host the CLI needs and what each is for.
2. The host strings in the **shipped Claude Code 2.1.260 binary** on this machine (`strings` over the Mach-O). `code.claude.com` appears 2046 times, `platform.claude.com` 30, the `mcp-proxy…` host 4.

Both sources agree on the three hosts the issue named. Their roles:

- **`platform.claude.com`** — Console account authentication, and OAuth token exchange, refresh and revocation *for claude.ai accounts too*. So it is required for both account types, which is why the reported symptom was an API error partway through a session rather than a login failure.
- **`claude.com`** — sign-in opens a `claude.com` page that redirects to claude.ai; pre-approved WebFetch documentation lookups also land here.
- **the `mcp-proxy…` host** — MCP connectors from claude.ai, on by default for claude.ai-authenticated users.

## What actually changed

Two entries, not three: `claude.com` and the `mcp-proxy…` host. `claude.com` is listed bare, per this file's stated convention, and `is_domain_match` treats a bare entry as exact-or-subdomain — so it already admits `platform.claude.com` and `code.claude.com`. Adding `platform.claude.com` separately would be a redundant entry to keep in sync. The same rule means `claude.ai` already covered `downloads.claude.ai`, so that documented host needed nothing either. The doc comment spells all of this out so the next reader does not "fix" the apparent omission.

**Deliberately left out**, all documented but none on the path to a working terminal session — each named on the const with its reason:

| Host | Why not |
| --- | --- |
| `storage.googleapis.com` | Plugin metadata for `/plugin`. Far too broad a host to grant by default — it is every Google Cloud Storage bucket. |
| `raw.githubusercontent.com` | `/release-notes` changelog feed. |
| `formulae.brew.sh` | Update check, Homebrew installs only. |
| The two Datadog intake hosts | Optional telemetry, which is what `blocked-domains.txt` exists to strip. |
| `bridge.claudeusercontent.com`, `*.frame.claudeusercontent.com` | Claude in Chrome and artifacts — not CLI-under-cplt paths. |
| `registry.npmjs.org` | Already in `PACKAGE_REGISTRY_DOMAINS`. |

Anyone who wants one of these adds it via `allowed_domains`.

## What I could not verify

The inherited **`statsig…` entry** could be corroborated from neither source: the docs attribute feature-flag fetches to the main API host, and the 2.1.260 binary carries no statsig host at all. I have left it in — removing it can only break a session, and this PR is meant to be additive — but annotated it as a deletion candidate once someone captures a run with `--observe-domains`.

I also did **not** run a live `--proxy-log-level all` session to collect `BLOCKED-ALLOWLIST` lines, which is the strongest evidence the issue asked for. The binary-string check is a weaker substitute: it proves a host is referenced by the shipped code, not that a given session reaches it.

The issue's wider suggestion — auditing every agent's list, or "last verified" dates, or #144 — is out of scope here. This const now carries a dated provenance note, which is the per-const version of that idea; the other agents' lists still do not.

## Scope of the fix

`proxy.default_allowlist` is **off by default**; only `--preset strict` turns it on. So this fixes strict-mode users, not everyone — an unmodified default install was never blocked by any of this.

Adding the `mcp-proxy…` host widens what a claude.ai MCP connector can reach, which is the #62 interaction the issue flags. The doc comment says so and names the two settings (`ENABLE_CLAUDEAI_MCP_SERVERS=false`, `disableClaudeAiConnectors`) that turn the connectors off for anyone who does not want that trade.

## Tests

`claude_allowlist_admits_the_hosts_claude_code_needs` asserts through `proxy::is_domain_match` — the matcher the proxy actually uses — not against the const. It checks that all seven hosts (including the three reached only by the subdomain rule) are admitted, that `evil.com`, `storage.googleapis.com` and the Datadog intake stay blocked, and that no entry contains a literal `*`.

**Mutation check:** deleting the two new entries fails it at `claude allowlist must admit claude.com`. Restored, it passes.

## Gates

`cargo fmt --check`, `cargo test --lib` (956), `cargo test --test unit_tests` (331), `cargo clippy --all-targets -- -D warnings` — all clean.

`docs/proxy.md`'s per-agent summary is updated to match.


Starefossen:
Both threads addressed. The console host is now documented alongside the statsig one, in a block that says plainly that neither could be corroborated against Anthropic's current requirements table or the 2.1.260 binary — the table attributes Console authentication to `platform.claude.com`, and neither host appears in the binary. Both are kept because removing an entry from a fail-closed list can only break a session, and both are flagged as deletion candidates pending an observed run. `docs/proxy.md` says the same. (This repo's publishing guard rejects the literal Anthropic domain string, hence the roundabout naming.)

PR #354, opened 2026-09-04
test(sandbox): pin the cplt exec grant overlap wiring

Starefossen:
Two non-blocking gaps from the review of #351. Stacked on it; retargets to `main` when #351 merges.

## The predicate was pinned, its wiring was not

#351's four tests for `writable_non_exec_tree_over` each exercise a distinct branch of the conjunction, and making the predicate return `None` fails one of them. But replacing the call in `generate_policy` with `writable_non_exec_tree_over(&[], &cplt_bin)` — the check entirely bypassed — passed all five. The helper was tested; the thing that calls it was not.

`generate_policy_skips_the_cplt_exec_grant_under_a_writable_tree` makes the test binary's own parent an `allow.write` tree (read+write, no execute — the pair the check exists to refuse) and asserts no rule for the binary is emitted at all.

Mutation-checked both ways:

- call replaced with `writable_non_exec_tree_over(&[], &cplt_bin)` → the new test fails, the other 60 in the module pass
- restored → 951 lib tests pass

The doc comment on the sibling `generate_policy_grants_execute_on_the_cplt_binary` now says out loud that it fails if the *test* binary lives under `/tmp` (`CARGO_TARGET_DIR=/tmp/...`). That is the check working correctly, but it looks like a broken test to whoever hits it.

## A stale comment contradicted the argument below it

The paragraph retained from #348 said `install.sh`'s default `~/.local/bin` "did not" re-enter, two paragraphs above the new argument — and SECURITY.md — saying `~/.local/bin` is read-only and already executable. The latter is right: it is `write: false, process_exec: true` in `HOME_TOOL_DIRS` and has been since #46.

Checked whether #348's premise being wrong undermines the grant itself. It does not, but the accident-of-layout story needed different names. Already covered by something else: `/usr/local`, `/snap`, `/home/linuxbrew/.linuxbrew` (`LINUX_TOOL_DIRS`), `~/.local/bin`, `~/.cargo/bin`, `go/bin` (`HOME_TOOL_DIRS`). Not covered by anything: `~/bin`, which `install.sh` picks when it is the writable directory on `PATH`, and whatever `--dir` points at. So the wrapper really did die for some installs, just not the default one — and the grant's actual justification is the sentence that was already there: the guard's design requires it, and leaving it to depend on install location is what made it look like it worked.

## Not done

`generate_policy` runs twice per launch (`main.rs` for the `cplt check` model, `sandbox.rs` for the real policy), so the skip warning prints twice. Deduplicating needs either process-global state inside a pure policy function — which makes it order-dependent under test — or a `quiet` flag threaded through the signature. Both are worse than the duplicate line. Left alone.

## Gates

`cargo fmt --check`, `cargo test --lib` (951), `cargo test --test unit_tests` (331), `cargo clippy --all-targets -- -D warnings` — all clean.

## Also picked up the two open reviewer threads on #351

The Copilot reviewer flagged the same `~/.local/bin` comment independently, plus one more of the same shape: `generate_policy_grants_execute_on_the_cplt_binary`'s failure message promises read+execute but the assertion only checked `execute` and `!write`, so dropping `read` would leave the test green and contradicting itself. It now checks `r.access.read` too.


Starefossen:
Reviewed independently before merge, against a checkout merged with current `main` rather than the branch alone. The CI runs on this PR were from 2026-09-05 and `main` has moved about twenty commits since.

**Gates on the merge result:** `cargo fmt --check`, `clippy --all-targets -- -D warnings`, 1017 lib tests, 344 unit tests. Clean.

**The claimed mutation still kills.** Replacing the call with `writable_non_exec_tree_over(&[], &cplt_bin)` fails exactly `generate_policy_skips_the_cplt_exec_grant_under_a_writable_tree` and nothing else.

**Every factual claim in the rewritten comment was verified against the code**, since the point of this PR is that the previous comment stated something false. `.local/bin` and `.cargo/bin` are `write: false, process_exec: true` in `HOME_TOOL_DIRS`; `/usr/local`, `/snap` and `/home/linuxbrew/.linuxbrew` are in `LINUX_TOOL_DIRS`; `~/bin` appears in no table, and `install.sh:111` really can select it. The corrected story holds.

## One gap found, and fixed here

A second opinion attacked the new test with five further mutations, all of which it killed. Two survived, both rewiring the check to a different source of truth:

- checking `config.extra_write` instead of the emitted rules
- checking `super::writable_trees(config)` instead of the emitted rules

The second one matters. The comment at `sandbox_landlock.rs:984` argues specifically that the check must run against the rules actually emitted, *because* the project dir is writable and executable by design and skipping the grant there would break the guard to close a hole already open by construction. Nothing tested that. I reproduced it: with the check rewired to `writable_trees`, **all 1016 tests pass** while the guard is silently disarmed for anyone running cplt from their own checkout.

That is the same defect shape this PR exists to fix — a property the comment asserts and no test pins — so it belongs here rather than in a follow-up. Added `generate_policy_keeps_the_cplt_exec_grant_under_the_project_dir`, which fails under that rewiring and passes on the clean tree.

The `extra_write` variant cannot be killed without making the binary path injectable, since a test cannot place `current_exe()` under `/tmp`. Not worth a signature change; worth knowing if the check is ever refactored.

## Not changed

The duplicated skip warning is left alone, and the call is right, though the stated reason is not the strongest one: a `OnceLock` at the warn site would be about three lines and nothing observes `ui::warn` under test. The better argument is that this is an error path where every guarded command is about to fail loudly anyway, and the structural fix — having `prepare` reuse the policy `main.rs` already built — needs a cfg split because macOS still builds its own model.

One pre-existing issue noted for follow-up, out of scope here: `generate_policy` is also called on macOS, where the warning's text ("every guarded git and gh command fails with Permission denied") is false, since macOS grants `process-exec` profile-wide. It only fires if cplt sits under `/tmp`, `~/.cache` or an `allow.write` tree.

Issue #126, opened 2026-07-05
Harden the test harness: vacuous, silent-skip, and untested-enforcement gaps


Starefossen:
# PRD: Harden the test harness — vacuous, silent-skip, and untested-enforcement gaps

## Problem

An adversarial audit of the test suite (not "do tests pass" but "would they catch a regression?") found a recurring class of tests that ship green while proving little. This is the same class already fixed piecemeal this cycle (the vacuous proxy-forced smoke test, the securityfs-vs-syscall probe, the invalid `-C` flag, the false-premise `--print-profile` pin). This issue tracks the remaining instances so a real security regression can't ship green.

### Tier 1 — a security regression would ship green (fix first)
- **macOS credential-dir read-deny has no runtime test that runs in CI.** `tests/integration.rs:208` (`sandbox_blocks_ssh_read`/`_kube`/`_docker`/`_aws`) build a **hand-written** SBPL profile (`write_test_profile`), not cplt's `generate_profile`, and self-skip when `~/.ssh`/`~/.aws` are absent (always, on a fresh runner). `e2e_projects.rs:617` self-skips the same way. So dropping the `~/.ssh` deny from cplt's real profile ships green on macOS. Fix: a runtime test that applies cplt's *real* generated profile with **planted** secrets in a temp HOME, executed under `CPLT_TEST_REQUIRE_SANDBOX`.
- **SBPL last-match-wins deny ordering has zero tests** (`sandbox_profile.rs:1358` test module). The whole macOS model relies on every `(deny …)` being emitted *after* its broader allow (the code comments repeatedly warn of this). Nothing asserts ordering; reorder a deny before its allow (silently permitting `.git/config` writes / credential reads) and all tests pass. Fix: assert deny-after-allow for the sensitive rules, ideally a runtime deny test.

### Tier 2 — vacuous / tautological (make them actually test, or delete)
- `unit_tests.rs:397` `allow_private_domains_bypasses_private_ip_check` — re-implements the decision with constants, never calls `handle_connect`; both asserts are constant-true. Deleting the real SSRF guard passes.
- `unit_tests.rs:2358` `allow_localhost_any_affects_both_backends` — always-true via a `|| p.contains("network-outbound")` disjunct (every profile contains that substring).
- `e2e_projects.rs:2475` `tcp_wildcard_bind` — asserts OK whether the 0.0.0.0 bind is denied **or** allowed.
[...]

PR #358, opened 2026-09-04
test: make gate refusals assert the reason, not just failure

Starefossen:
Closes the Tier 3 items of #126 that are about **vacuous negative assertions** — a check that reads as present and is not.

## The class of defect

`assert!(!ok, "…")` passes when the command failed for *any* reason: a renamed flag clap rejects, a fixture that never got built, a panic in the harness. Nine defects found in one review round were this shape, and in every one the helper was tested while the call site was not.

## What changed

**`tests/common/mod.rs`** — one new helper:

```rust
assert_refused(&stderr, ok, "gh api graphql allows arbitrary mutations");
```

It asserts the command failed, that it failed as a policy refusal (`BLOCKED by sandbox`), and that the refusal names the given reason. `assert!(!ok, …)` no longer appears anywhere in `tests/e2e_guards.rs`, so the convention is greppable.

**`tests/e2e_guards.rs`** — all 58 negative gate assertions converted, each pinned to the gate's own `Reason:` line (probed from the real binary, not guessed). Net −11 lines.

Two were weaker than they looked and are now honest:

- `gh_gate_blocks_graphql_endpoint` asserted `stderr.contains("graphql")` — satisfied by the echoed command line for *any* refusal.
- `gh_gate_blocks_cross_repo_pr_close` said "should mention scope" while asserting only the generic banner. Its sibling `gh_gate_blocks_pr_merge_cross_repo` is now pinned to `merging is a human decision`, because `pr merge` is `Decision::Block` outright — the scope check never runs there, and the old name implied otherwise.

**`tests/integration_linux.rs`** — the two named in #126:

- `proxy_blocks_curl_to_localhost_without_allow_flag` bound an ephemeral port and **dropped the listener** before running curl, so ECONNREFUSED passed it whether or not anything was enforced. Merged with the standalone allow-flag test it duplicated into `proxy_localhost_reachable_only_with_allow_flag`: both halves run against one live listener, positive control in the same test.
- `bwrap_cannot_see_host_init` keyed on the literal `"systemd"`, passing on any host whose init has another name. Now compares the sandboxed `/proc/1/comm` against the host's real one.

**`.github/workflows/ci.yaml`** — `cargo test --no-fail-fast` in all three test steps. cargo stops at the first failing *target*, so a broken lib assertion hides every integration test behind it; that hid two mutation-testing rounds in one day. Failures still fail the job.

## Mutation evidence

Injecting a wrong-reason failure into the gh harness (an unrecognised flag, so cplt exits non-zero before any policy runs):

| | passed | failed |
|---|---|---|
| before | 82 | 41 |
| after | 65 | 58 |

17 tests that previously shipped green over a gate that never ran now go red. The Linux pair is verified in a separate red CI run (see comment below).

## Not done, deliberately

- **Silent-skip visibility** — already built. `require_sandbox!`/`require_landlock!`/`require_bwrap!`/`require_curl!` panic under `CPLT_TEST_REQUIRE_SANDBOX=1`, and the macOS job counts residual skip markers and fails on unexpected ones.
- **A positive-control framework** — rejected. Where the refusal text distinguishes the cause, pinning it is the cheaper proof; a control only earns its place where the message cannot tell the intended block from a blanket one (the `protect_default_branch_only` tests, whose controls are the adjacent `_allows_` tests in the same file).

Refs #126

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -53,7 +53,11 @@ jobs:
         run: cargo clippy --all-targets -- -D warnings
 
       - name: Run unit and lib tests
-        run: cargo test --test unit_tests --lib --verbose
+        # --no-fail-fast: cargo stops at the first *target* that fails, so a
+        # broken lib assertion hides every integration test behind it. Two
+        # mutation-testing rounds were scored against suites that never ran for
+        # exactly that reason (issue #126). Failures still fail the job.
+        run: cargo test --no-fail-fast --test unit_tests --lib --verbose
 
   # Integration tests need macOS for sandbox-exec
   integration-tests:
@@ -140,6 +144,10 @@ jobs:
           dotnet-version: '8.0.x'
 
       - name: Run all tests (skips fail the job)
+        # --no-fail-fast: cargo stops at the first *target* that fails, so a
+        # broken lib assertion hides every integration test behind it. Two
+        # mutation-testing rounds were scored against suites that never ran for
+        # exactly that reason (issue #126). Failures still fail the job.
         # CPLT_TEST_REQUIRE_SANDBOX=1 makes require_sandbox! panic instead of skip,
         # so a missing/broken sandbox-exec (a Seatbelt regression on the runner)
         # turns this job red rather than silently passing while nothing is
@@ -153,7 +161,7 @@ jobs:
         shell: bash
         run: |
           set -o pipefail
-          cargo test --verbose -- --nocapture 2>&1 | tee "$RUNNER_TEMP/test-output.log"
+          cargo test --no-fail-fast --verbose -- --nocapture 2>&1 | tee "$RUNNER_TEMP/test-output.log"
 
       - name: Surface tests skipped for missing tooling
         if: always()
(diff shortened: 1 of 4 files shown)


Starefossen:
## Mutation evidence for the two Linux tests

Run on the throwaway branch `test/mutation-proof-126` (#359, now closed): https://github.com/navikt/cplt/actions/runs/33921973061

| test | mutation | result |
|---|---|---|
| `proxy_localhost_reachable_only_with_allow_flag` | negative half runs **with** `--allow-localhost`, i.e. loopback enforcement off | **FAILED** |
| `mutation_control_old_localhost_test_is_vacuous` (the pre-PR version, verbatim) | same mutation | **ok** — passes with enforcement fully disabled |
| `bwrap_cannot_see_host_init` | sandboxed PID 1 reports the host's init name, i.e. the PID namespace does not isolate | **FAILED** |

```
test linux_tests::bwrap_cannot_see_host_init ... FAILED
test linux_tests::mutation_control_old_localhost_test_is_vacuous ... ok
test linux_tests::proxy_localhost_reachable_only_with_allow_flag ... FAILED
test result: FAILED. 76 passed; 2 failed
```

The middle row is the point: the test this PR replaces passes while the thing it claims to check is switched off.

### One mutation that did not model the defect

The first attempt disabled the PID namespace with `--no-bubblewrap` and `bwrap_cannot_see_host_init` still passed (https://github.com/navikt/cplt/actions/runs/33921759881). Not a hole in the strengthened assertion — on that path `/proc/1/comm` is not readable at all, so the script falls through to `echo 'blocked'`, which both the old and new forms accept. Modelling the defect directly (PID 1 reads back as the host init) turns it red.

### Where the old `bwrap` assertion is vacuous, and why CI cannot show it

`!stdout.contains("systemd") || stdout.contains("blocked")` fails on a systemd runner even with the defect present, so a GitHub runner cannot distinguish the two forms. Its vacuity is specific to a host whose init is named otherwise — a container with `tini`, a non-systemd distro — where `!contains("systemd")` is true and the assertion holds no matter what PID 1 is. That is by inspection, not by CI run.

Issue #394, opened 2026-09-07
Advisories and release notes owed for the 2026-09-06 security fixes


Starefossen:
Seven security fixes merged on 2026-09-06 and 2026-09-07. None of them has an advisory, and the newest published advisory is patched at `2026.09.06-082853-ce62f19`, which predates all of them. The decision on the second audit was to fix in public PRs and publish an advisory naming the fixed version, so this is the unfinished half of that.

## Advisories owed

| PR | Fix | Advisory |
|---|---|---|
| #380 | TIOCSTI terminal injection was not denied on Linux (C-01) | none |
| #381 | An explicit deny path was re-allowed by an overlapping docker/gpg grant | none |
| #382 | Push authorization was not bound to the real push destination (H-10) | none |
| #384 | Pi's trust store was writable, and managed binaries sat inside writable grants | none |
| #391 | A shim whose target is outside every grant failed with exit 126 instead of naming it | none, likely not security-reportable |

#379 and #377 are already covered by GHSA-4hq4-fc3p-7gv2 and GHSA-q67f-29jw-3xqr.

Credit for the reported findings goes to @randax.

## Prerequisite

Each advisory needs a released version to name. Publishing one with an empty `patched_versions` is the honest answer only when nothing fixes it; here the fix exists and the tag does not yet. See #370 — a red `main` silently skips Release, which is how an advisory ended up with no patched version before.

So the order is: cut the release carrying these commits, then publish, then set the patched version. Not the other way round.

## Release notes owed

#382 carries two breaking changes that no release note describes:

- A rule constrained by `branches` now rejects `git push --all`, because the destination set cannot be enumerated and the guard fails closed.
- A mistyped `allow_push` key now fails the launch instead of being ignored.

Both are deliberate and both will surprise someone mid-session. They need to be in the release notes for the version that carries them, not discovered from a refusal message.

## Done when

- The release carrying these commits is tagged.
- An advisory exists for each entry above that warrants one, naming that version, and carrying the "not fixed by this" section.
- The release notes state the two breaking changes.

Starefossen:
**The release exists: `2026.09.07-083818-afc0751`.** Verified by ancestry rather than by timing — every PR listed above is an ancestor of that tag:

```
#380 IN tag   #381 IN tag   #382 IN tag
#384 IN tag   #391 IN tag   #387 IN tag
```

So the prerequisite is met and the advisories can name a patched version.

[...]

PR #399, opened 2026-09-07
fix(gh-guard): allow the gh credential helper so HTTPS push works

### LINKED ISSUE #396: gh gate blocks 'gh auth git-credential', breaking HTTPS push inside the sandbox
## Summary

The gh policy table has no entry for `gh auth git-credential`, so the credential helper that `gh auth setup-git` installs is denied by the default-deny path for unknown commands. Inside the sandbox this removes the credential helper for HTTPS pushes: git gets no username or password and fails.

## Reproduction

Directly against the gate:

```
$ printf 'protocol=https\nhost=github.com\n\n' | cplt gh-gate \
    --real-gh "$(command -v gh)" --unknown-command=block -- auth git-credential get
⚠️ BLOCKED by sandbox: 'gh auth git-credential' is not recognized by the policy table.
This command may have been added in a newer gh CLI version.
This operation is restricted by the cplt sandbox (default-deny for unknown commands).
```

End to end inside a sandbox, with `credential.https://github.com.helper = !gh auth git-credential` in `~/.gitconfig` (what `gh auth setup-git` writes):

```
$ cplt --project-dir <clone> exec -- git push --dry-run origin HEAD:refs/heads/x
⚠️ BLOCKED by sandbox: 'gh auth git-credential' is not recognized by the policy table.
fatal: could not read Username for 'https://github.com': terminal prompts disabled
```

## Why it happens

`POLICY` in `src/gh_proxy.rs` carries `auth` entries for `status`, `token`, `login`, `logout`, `refresh`, `setup-git` and `switch` (src/gh_proxy.rs:645-687). `git-credential` is absent, so `evaluate_with_policy` returns `Decision::Unknown` and the refusal at src/gh_proxy.rs:1914 fires under the default `unknown_command: Block` (src/config/types.rs:713).

The shim is genuinely on the path that git takes: `~/.gitconfig` is readable inside the sandbox (src/sandbox_profile.rs:685), the only `GIT_CONFIG_*` overrides applied to the agent's git are `commit.gpgsign` and `tag.gpgsign` (src/sandbox_policy.rs:736-762), the `credential.helper` neutralisation in src/git.rs:367-371 is parent-side only, and `$SCRATCH/bin` is prepended to PATH (src/sandbox_exec.rs:624-633).

## Impact

Depends on the agent, which is worth stating explicitly because the blast radius differs:

- Agents granted Keychain access (Copilot, Antigravity, Claude, Goose — src/agent.rs:416-421, src/sandbox_profile.rs:496-505) can still get a credential by invoking the real `gh` binary by absolute path, so for them this is a broken documented workflow rather than a hard block.
- OpenCode, Pi and plain `cplt exec` have no credential channel at all. Verified inside `cplt exec`: `security find-generic-password -s gh:github.com` exits 44 (exits 0 outside the sandbox), `gh auth git-credential get` returns empty with exit 1, and `gh api user` reports "Requires authentication".

Two related gaps make this harder to work around or diagnose:

- `GH_TOKEN`, `GITHUB_TOKEN` and `COPILOT_GITHUB_TOKEN` are stripped for every non-Copilot agent (src/sandbox_env.rs:27-46), so the "export GH_TOKEN yourself" hint at src/sandbox_exec.rs:296 does not help those agents.
- `git credential fill` is refused by the git gate's subcommand allowlist ("not a recognized subcommand"), so an agent cannot inspect the credential chain to work out what failed.

## Suggested fix

Add the missing policy entry beside the other `auth` rows:

```rust
PolicyEntry {
    command: "auth",
    subcommand: "git-credential",
    decision: Decision::Allow,
    reason: "git credential helper — required for HTTPS push",
},
```

Open question worth deciding explicitly: `gh auth git-credential get` prints `password=<token>` on stdout, so allowing it unconditionally adds a second gh verb that reveals the token while `gh auth token` and `gh auth status --show-token` stay gated by `block_auth_token` (src/gh_proxy.rs:1785-1790). The marginal exposure is close to zero for Keychain-granted agents, which can already read the token with `security find-generic-password`, and SECURITY.md:440 already states that `block_auth_token` is not a same-UID confidentiality boundary. Still, either gate `git-credential` under `block_auth_token` as well, or document why it is treated differently.

A regression test belongs next to `default_deny_unknown_commands` (src/gh_proxy.rs:3368), asserting that `auth git-credential` is allowed.

### LINKED ISSUE #397: docs: known-impacts recommends a credential helper the gh gate blocks
## Summary

`docs/known-impacts.md` tells operators to set up the exact credential helper that the gh shim then denies, and claims a credential channel that only exists for Copilot.

## What the docs say

- `docs/known-impacts.md:292`: "Git commit and push **work out of the box** over HTTPS, no extra flags needed."
- `docs/known-impacts.md:310-312`: recommends `gh auth setup-git`, which writes `credential.https://github.com.helper = !gh auth git-credential`.
- `docs/known-impacts.md:331`: "`git fetch/pull/push` (HTTPS) ✅ Works — Port 443 allowed, `gh auth token` provides credentials".

## Why each is wrong

The recommended helper is blocked by the gh policy table, which has no `auth git-credential` entry (see the separate issue for that bug). This affects Copilot too: its `git push` goes through the same helper, and `GH_TOKEN` is not in its environment by default since `inject_token` defaults to false (src/config/types.rs:712).

`gh auth token` inside the sandbox is served only from the scratch cache, and that cache is written for Copilot only (`should_cache_token`, src/sandbox_exec.rs:411-413; served at src/main.rs:3106-3130). For any other agent it prints "No cached token available", so the stated mechanism does not exist for them.

An additional omission: `GH_TOKEN`, `GITHUB_TOKEN` and `COPILOT_GITHUB_TOKEN` are stripped for non-Copilot agents (src/sandbox_env.rs:27-46). Any doc or hint suggesting the operator export a token themselves is dead advice for those agents.

## Suggested fix

Once the policy entry lands, rewrite the section to state per agent what actually works, and correct `:331` to name the credential helper rather than `gh auth token`. Also worth documenting that `gh pr create` in a non-TTY context refuses to push at all ("you must first push the current branch to a remote, or use the --head flag"), so the branch must be pushed first.

Starefossen:
Fixes #396. Fixes #397.

`gh auth setup-git` installs `credential.https://github.com.helper = !gh auth git-credential`, but the policy table had no entry for that verb, so it fell through to the default-deny path for unknown commands. Every HTTPS push inside the sandbox failed:

```
⚠️ BLOCKED by sandbox: 'gh auth git-credential' is not recognized by the policy table.
fatal: could not read Username for 'https://github.com': terminal prompts disabled
```

Reported from a real session in nais/unleash, where it blocked both a normal push and `--force-with-lease`.

## The entry is ungated

`git-credential get` prints `password=<token>`, so allowing it unconditionally adds a second gh verb that reveals the token while `gh auth token` stays intercepted. That is deliberate:

- Gating it behind `block_auth_token` (default true) would leave HTTPS push broken in the default configuration, which is the bug.
- It withholds nothing. Agents with a Keychain grant (copilot, antigravity, claude, goose) can already read the token with `security find-generic-password -s gh:github.com -w`; opencode, pi and `cplt exec` have no token source at all, so the helper prints nothing for them.
- `block_auth_token` is already documented as best-effort rather than a same-UID boundary (SECURITY.md, honest gaps).

`store` and `erase` are allowed along with `get` because gh's helper implements both as no-ops (`pkg/cmd/auth/gitcredential/helper.go`: "we pretend to implement the 'store' operation, but do nothing"), so nothing durable is written or removed.

## Docs

`docs/known-impacts.md` claimed push "works out of the box" and credited `gh auth token` for the credentials. Both were wrong: the recommended helper was the denied one, and the `gh auth token` cache is written for Copilot only (`should_cache_token`). The section now states what each agent actually has, marks the Keychain table macOS-only, notes that `keychain_substitute` can drop the grant for claude/antigravity, and names `--pass-env GH_TOKEN` as the explicit opt-in.

Two threat-model rows (`SECURITY.md`, `docs/gh-guard.md`) claimed the raw token stays unreachable. They now name this exception.

## Tests

- `auth_git_credential_allowed` — the policy decision and the full `gate` with `block_auth_token` on, asserting `auth token` stays gated.
- `git_credential_helper_reaches_the_real_binary` — the shim's actual entry point (`decide_gh_gate`), asserting the helper execs rather than hitting the token-cache intercept.

## Verification

```
$ printf 'protocol=https\nhost=github.com\n\n' | ./target/debug/cplt gh-gate \
    --real-gh "$(command -v gh)" --unknown-command=block -- auth git-credential get
protocol=https
host=github.com
username=Starefossen
password=gho_<redacted>
```

`cargo fmt`, `cargo clippy --all-targets -- -D warnings` and `cargo test` all clean.


Starefossen:
Flagging this after the fact, for the record and for whoever picks the work up next. This is not a request to revert.

This PR narrows a documented security property, and it merged without a human deciding that explicitly. `gh auth git-credential` is allowed ungated, and the verification block in the description shows what that means:

```
password=gho_<redacted>
```

So there are now two gh verbs that reveal the raw token, while `block_auth_token` continues to intercept `gh auth token`. The PR is straightforward about this and updates the two threat-model rows that previously claimed the raw token stays unreachable, which is the right thing to have done.

The reasoning holds: gating the helper behind `block_auth_token` would leave HTTPS push broken in the default configuration, and HTTPS push was genuinely broken inside the sandbox (#396, reported from a real session in nais/unleash). Given a choice between a control that is bypassed by everyone who needs to push and one that is honest about its limit, this is the better trade.

What is worth a deliberate confirmation rather than silence is the *shape* of the change: `block_auth_token`'s premise is that the token is not readable from inside the sandbox, and that premise is now narrower than its name suggests. Anyone reasoning about the threat model from the flag name alone will get it wrong.

Two things worth doing on the back of this, neither blocking:

- Consider whether `block_auth_token` still describes what it does, or whether the exception belongs in the flag's own help text and not only in the threat-model tables.
- The next advisory or release note touching token exposure should state this exception plainly, in the same way #394 requires a "not fixed by this" section.

Tracked in the handoff at #405.

Issue #431, opened 2026-09-08
Golden-path tests: the suite checks the agent's view of the sandbox, never the operator's view of cplt


Starefossen:
An assessment of the end-to-end and golden-path coverage, prompted by five defects in one week that passed every unit test and were caught by a human running the tool:

- an applied per-repo local config left no trace on any surface while its grants were in effect
- `config show` reported both guards `enabled = false` while they were enforcing (#410)
- `config set --local sandbox.repo_dirs /nonexistent` succeeded and then failed every later launch
- `cplt settings` read the local file with a different loader than the launch, so a stale file showed as applied in the TUI while the launch ignored it
- the git guard's default was `warn` while `config show` said `block`, and three docs disagreed again

## The finding

**The gap is not "not enough e2e". The e2e tests assert on the agent's view of the sandbox and never on the operator's view of cplt.**

57 tests prove the sandbox blocks `~/.ssh`. Zero prove the startup summary tells the truth about the guards. Grepping the whole test tree for `gh guard:`, `git guard:`, `WARN MODE` or `Command guards` returns nothing. Every one of the five defects lived on stderr or in `config show`, and the suite reads those surfaces in about four places.

Where the binary does run, assertions are overwhelmingly on the *child's* stdout (the `RESULT:` protocol in `e2e_projects.rs`) and on exit codes. stderr is captured everywhere and asserted on almost nowhere — which is where the summary, the `(local)` labels and the guard lines all live.

A second structural fact: **the 150 gate tests never pass through config.** `e2e_guards.rs` calls `cplt gh-gate` / `cplt git-gate` directly with policy flags on the command line, so config resolution, wrapper generation and the launch are all bypassed. They are thorough about `git push` spellings and prove nothing about whether a given config wires that policy into the wrapper. Only two tests in the tree go config → launch → generated wrapper → gate.

## Cost is not the problem

The entire suite is ~55 s locally. A golden-path file of 10–15 fake-agent launches adds 5–10 s. The `integration-tests` CI job is 3m20s, of which ~2 min is toolchain, Chrome, Java and .NET setup rather than tests.

[...]

PR #439, opened 2026-09-08
fix(check): report what will actually happen, not what the policy objects to

Starefossen:
## The defect

`cplt check exec <cmd>` exists so a developer can ask what will happen without launching an agent. That makes it the one surface where telling the truth about the guards *is* the feature. It got warn mode backwards.

`gate_git` and `gate` answer one question — does the policy object? — and return `Err` when it does. Whether that objection *stops* the command is a separate decision the launch makes from the enforcement mode, and `check exec` skipped it. Every objection was reported `BLOCKED`.

Observed, with `git_guard.mode = "warn"`:

```
$ cplt check exec git push origin main
git push origin main: BLOCKED

$ cplt exec -- git push origin main
⚠️  WARNING (would block): ...
   751c4c3..92e9978  main -> main      # the push ran, the remote moved
```

Both guards had it: the gh arm builds a `GatePolicy` carrying the mode, but `gate` refuses without consulting it — the mode is applied later, by `decide_gh_gate`.

Warn and audit now report the command allowed, with the objection in the reason and the fix naming how to make it bite. The block-mode message is stripped of its `BLOCKED by sandbox` prefix on that path, for the same reason `warn_mode_message` already strips it in the launch: a line that says `ALLOWED` must not have `BLOCKED` nested inside it.

```
git push origin main: ALLOWED
  Reason: the git guard objects, but it runs in warn mode, so the command runs. 'git push origin' is not allowed while push prevention is active.
  Fix: set git_guard.mode = "block" to enforce this.
```

## Golden paths 5–7 of #431

Each falsified before being trusted.

**Path 5** pins the above end to end: `check exec` and a real push against a bare origin, in both modes, required to agree twice over — blocked with the ref unmoved, then allowed with the ref moved. Restoring `Decision::Blocked` fails it.

**Path 6** is the repo layer's trust model: an untrusted `.cplt.toml` must be announced in the summary, inert until approved, listed pending by `cplt trust`, and in force after `trust accept --all`. Making the approval check pass unconditionally fails it.

Worth recording: my first falsification of path 6 mutated `trust::is_key_approved` and the test stayed green. That function is short-circuited by `is_some_and` when no trust file exists, so the mutation never ran — the test was fine, the falsification was aimed at the wrong function. The gate that decides is the `is_approved` closure in `apply_repo_config`. A falsification that silently tests nothing is how a dead test gets trusted, so it seemed worth naming rather than quietly redoing.

**Path 7** covers three ordinary repository states that were untested anywhere: no remote, detached HEAD, uncommitted changes. Each must launch and still report the guards. The guards read the remote's default branch, so "there isn't one" is a real input rather than an error.

Closes the #431 ranked list. Follow-up noticed while testing, not fixed here: with no remote at all the git guard fails closed and blocks every push, while the summary row still says "blocks git push to the default branch" — the same overstatement fixed in #436, in the other direction.

### DIFF
--- a/src/check.rs
+++ b/src/check.rs
@@ -490,6 +490,19 @@ pub struct ExecContext<'a> {
     pub gh_guard: &'a crate::config::GhGuardPolicy,
     pub git_guard: &'a crate::config::GitGuardPolicy,
     pub project_dir: &'a Path,
+    /// The repository facts the launch bakes in, captured the same way and from
+    /// the same trusted git.
+    ///
+    /// Without them `protect_default_branch_only` has no default branch to
+    /// compare against, so `gate_git` fails closed and every push reads as
+    /// blocked — including a feature-branch push the launch allows. That is
+    /// the most common push there is, so the surface built to answer "what
+    /// will happen" was wrong about the ordinary case.
+    pub repo_facts: &'a crate::gh_proxy::RepoFacts,
+    /// Both guards install themselves through a PATH shim in the scratch dir.
+    /// Without one there is no shim, so the guards do not run whatever the
+    /// config says — which is what the launch summary reports as `inactive`.
+    pub scratch_dir: bool,
 }
 
 /// Static explanation of whether a command would run under the resolved policy.
@@ -509,14 +522,6 @@ fn command_basename(cmd: &str) -> String {
         .unwrap_or_default()
 }
 
-/// Explain whether `argv` would run, is guard-blocked, or needs an `allow_*`.
-///
-/// Reuses the gh/git guard classifiers ([`crate::gh_proxy::gate`],
-/// [`crate::gh_proxy::gate_git`]) so the verdict matches what the in-sandbox
-/// wrappers enforce, and the `allow_docker` / `allow_tmp_exec` gates for the
-/// other high-frequency cases. Unrecognized commands get an honest generic
-/// answer (they run, subject to the fs/net policy).
-#[must_use]
 /// What a guard refusal means for `check exec`, given the mode that guard runs in.
 ///
 /// `gate_git` and `gate` answer one question — does the policy object? — and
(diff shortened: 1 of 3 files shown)


Starefossen:
Adversarial review found the fix incomplete: `check exec` was still wrong on the **default** config, in two more places, and the first commit's framing ("check exec now agrees with the launch") would have been read as more than it delivered. Both are now fixed, with the same root cause — the launch hands the guards inputs `check exec` did not.

**A feature-branch push read as `BLOCKED`.** The launch bakes repository facts at startup and passes a trusted git so the gate can resolve which branches a push targets. `check exec` passed neither, so `protect_default_branch_only` had no default branch to compare against and fell through to its fail-closed arm. Verified before and after:

```
# before
$ cplt check exec git push origin feature/x     → BLOCKED
$ cplt exec -- git push origin feature/x        → pushed, ref moved

# after
$ cplt check exec git push origin feature/x     → ALLOWED
$ cplt check exec git push origin main          → BLOCKED
```

The comment justifying the empty facts said they matched "what a launch that captured nothing would do" — but the launch always captures, so that launch does not exist.

**`sandbox.scratch_dir = false` read as `BLOCKED`.** Both guards install through a PATH shim in the scratch dir, so without one they do not run however they are configured — the launch summary already says `inactive`. `check exec` consulted only `enabled`. This is the worst direction for this surface to be wrong in: it tells an operator they are protected when they are not.

Test changes from the same review:

- the block half asserted `contains("BLOCKED")`, which the guard's quoted message satisfies whatever the verdict says — a mutation returning `Allowed` for block mode passed. Now pinned to the verdict line.
- the refusal half accepted any non-zero exit; now goes through `assert_refused`.
- warn mode must still print its notice, so a mutation that enforces nothing *silently* no longer passes.
- two new cases, each falsified by reverting its fix.

Also corrected: `refusal_decision` had been inserted between `explain_exec`'s doc comment and the function, silently transferring that comment and its `#[must_use]` to the helper.

And a claim in the first commit message was wrong — no-remote launches were **not** "untested anywhere"; `e2e_exec_default_branch_warning_only_when_actionable` covers one. Detached HEAD and dirty worktree are genuinely new.

Known and deliberately not fixed here, both pre-existing and worth their own issues: `gh auth token` reports blocked while the launch serves it from cache, and the objection in warn/audit is now only recoverable from prose rather than a structured JSON field.

PR #452, opened 2026-09-08
fix(pi): let Pi take its trust lock where the overlay can hold the root

### LINKED ISSUE #449: Pi: Error: EACCES: permission denied, mkdir '/home/mark/.pi/agent/trust.json.lock' when running cplt on a repository I've already trusted
Following the docs I've ran `/trust` with Pi outside cplt. I can see that `~/.pi/agent/trust.json` contains the repository path.

However, when I then run cplt I see the following error

```logs
Error: EACCES: permission denied, mkdir '/home/mark/.pi/agent/trust.json.lock'
    at Object.mkdirSync (node:fs:1370:26)
    at newFs.<computed> [as mkdir] (file:///usr/local/lib/node_modules/@earendil-works/pi-coding-agent/dist/bundle/chunks/chunk-WZB2R5YO.js:141:22123)
    at acquireLock (file:///usr/local/lib/node_modules/@earendil-works/pi-coding-agent/dist/bundle/chunks/chunk-WZB2R5YO.js:141:17129)
    at RetryOperation._fn (file:///usr/local/lib/node_modules/@earendil-works/pi-coding-agent/dist/bundle/chunks/chunk-WZB2R5YO.js:141:20170)
    at RetryOperation.attempt (file:///usr/local/lib/node_modules/@earendil-works/pi-coding-agent/dist/bundle/chunks/chunk-WZB2R5YO.js:141:9501)
    at file:///usr/local/lib/node_modules/@earendil-works/pi-coding-agent/dist/bundle/chunks/chunk-WZB2R5YO.js:141:20157
    at resolveCanonicalPath (file:///usr/local/lib/node_modules/@earendil-works/pi-coding-agent/dist/bundle/chunks/chunk-WZB2R5YO.js:141:16960)
    at lock (file:///usr/local/lib/node_modules/@earendil-works/pi-coding-agent/dist/bundle/chunks/chunk-WZB2R5YO.js:141:20021)
    at file:///usr/local/lib/node_modules/@earendil-works/pi-coding-agent/dist/bundle/chunks/chunk-WZB2R5YO.js:141:22469
    at AsyncFunction.lockSync (file:///usr/local/lib/node_modules/@earendil-works/pi-coding-agent/dist/bundle/chunks/chunk-WZB2R5YO.js:141:23280)
    at acquireTrustLockSync (file:///usr/local/lib/node_modules/@earendil-works/pi-coding-agent/dist/bundle/chunks/chunk-WZB2R5YO.js:1253:252)
    at withTrustFileLock (file:///usr/local/lib/node_modules/@earendil-works/pi-coding-agent/dist/bundle/chunks/chunk-WZB2R5YO.js:1253:671)
    at ProjectTrustStore.getEntry (file:///usr/local/lib/node_modules/@earendil-works/pi-coding-agent/dist/bundle/chunks/chunk-WZB2R5YO.js:1253:1535)
    at ProjectTrustStore.get (file:///usr/local/lib/node_modules/@earendil-works/pi-coding-agent/dist/bundle/chunks/chunk-WZB2R5YO.js:1253:1484)
    at resolveProjectTrusted (file:///usr/local/lib/node_modules/@earendil-works/pi-coding-agent/dist/bundle/chunks/chunk-WZB2R5YO.js:1278:1143)
    at async Object.resolveProjectTrust (file:///usr/local/lib/node_modules/@earendil-works/pi-coding-agent/dist/bundle/chunks/chunk-WZB2R5YO.js:1558:5829)
    at async DefaultResourceLoader.reload (file:///usr/local/lib/node_modules/@earendil-works/pi-coding-agent/dist/bundle/chunks/chunk-WZB2R5YO.js:1246:55351)
    at async createAgentSessionServices (file:///usr/local/lib/node_modules/@earendil-works/pi-coding-agent/dist/bundle/chunks/chunk-WZB2R5YO.js:1252:1833)
    at async createRuntime (file:///usr/local/lib/node_modules/@earendil-works/pi-coding-agent/dist/bundle/chunks/chunk-WZB2R5YO.js:1558:5528)
    at async createAgentSessionRuntime (file:///usr/local/lib/node_modules/@earendil-works/pi-coding-agent/dist/bundle/chunks/chunk-WZB2R5YO.js:1252:11569)
    at async main (file:///usr/local/lib/node_modules/@earendil-works/pi-coding-agent/dist/bundle/chunks/chunk-WZB2R5YO.js:1558:8556) {
  errno: -13,
  code: 'EACCES',
  syscall: 'mkdir',
  path: '/home/mark/.pi/agent/trust.json.lock'
```
It looks like Pi creates a lock file to read the trust file https://github.com/earendil-works/pi/blob/f53ac1135149f03fd1e2a5bfd29861120eaf5b96/packages/coding-agent/src/core/trust-manager.ts#L221 and this fails as cplt denies writes.

Running on WSL2 with bubblewrap enabled.

Starefossen:
Reported in #449 by an outside user: **Pi will not start under cplt on Linux.**

```
Error: EACCES: permission denied, mkdir '/home/mark/.pi/agent/trust.json.lock'
  at acquireTrustLockSync … ProjectTrustStore.getEntry … resolveProjectTrust
```

Pi takes a **write lock to read** its trust store, at startup, before a session exists. #384 narrowed `~/.pi/agent` to read-only precisely so `bin/` would not sit inside a writable tree — on Linux a writable parent unions into the exec-only child, and a contained agent could overwrite the managed `fd`/`rg` that the user's next unsandboxed `pi` runs.

The documented cost of that narrowing was *"Pi cannot create a NEW top-level entry under `~/.pi/agent` mid-session"*. `trust.json.lock` is exactly that — so the cost turned out to be fatal rather than inconvenient, for a user who had trusted the repository correctly outside cplt.

## The fix

Where bubblewrap is available, the narrowing is no longer the control. `bin/` is re-bound read-only inside the writable tree via `ro_protect`'s exec-only filter, and `trust.json`, `settings.json`, `extensions/`, `npm/` and `git/` through `host_persistence_paths` — the same protection macOS gets from profile-tail write denies. So the root is granted writable there and the overlay holds it.

Gated on bubblewrap being **available**, not on `use_bubblewrap` being requested: widening a grant on the strength of a request that then falls back to Landlock-only would remove the protection rather than relocate it.

## The part I nearly got wrong

`bin/` stays in the returned grants as an exec-only entry. That is not a leftover. `ro_protect_paths` selects `!write && process_exec` to decide what to re-bind, so dropping it would have granted the writable root and taken nothing back — reopening H-13/H-05 *through the fix for this issue*.

My first version did exactly that. The existing test asserting "nothing writable may contain `bin/`" is what caught it, which is a good argument for that test's phrasing — it pinned the property rather than the list.

That invariant is now asserted per regime: unchanged without bubblewrap, and with it the assertion becomes the thing that replaces it — `bin/` still listed exec-only, which is what puts it in the overlay.

## What is still broken

**Linux without bubblewrap is unchanged, and Pi still cannot start there.** There is no mechanism to take a write back inside a Landlock-granted tree, so that is the honest state rather than something to paper over by widening blindly. `known-impacts.md` now says so, and names installing bubblewrap or running Pi outside cplt as the options.

Worth raising upstream separately: taking a write lock to read a trust file is arguably Pi's bug, and fixing it there would remove the need for any of this.

Closes #449.


Starefossen:
Holding this rather than merging, and it will likely be superseded.

Two problems with the approach, the second found only after opening it.

**It probably does not fix the reported case.** The reporter said bubblewrap was *enabled*, meaning configured. Bubblewrap derives its writable binds from the same `fs_rules` as Landlock, so a read-only `~/.pi/agent` is bound read-only and `mkdir` there fails with `EROFS`. They got `EACCES`, which is Landlock's denial code. That points to bubblewrap being configured but not active — common on WSL2, where unprivileged user namespaces are often restricted and auto-detect falls back silently. If so, a bubblewrap-gated fix does nothing for them.

**It makes the overlay the sole control, and Linux CI said so.** `pi_managed_binaries_are_not_inside_a_writable_linux_grant` asserts at the Landlock rule level that nothing may make `~/.pi/agent/bin` writable. Granting the root writable breaks that: Landlock unions ancestors and cannot subtract, so only the read-only re-bind holds `bin/`. And bubblewrap silently skips a re-bind whose source does not exist — `bin/` does not exist until Pi downloads `fd`/`rg`, so on a fresh install the protection would not be applied at all. My first version also dropped `bin/` from the grant list entirely, which would have removed it from the overlay too; the test caught that.

**The direction instead**, per the maintainer: handle this in the model rather than as a per-agent exception. `AgentDir` already has `write_files` for narrow carve-outs in a read-only tree, honoured by both backends — but it grants write on paths that *exist*, and creating an entry is a right on the parent, so it cannot express a lock file. The missing primitive is a create-and-remove capability on an otherwise read-only directory: `MakeDir | MakeReg | RemoveDir | RemoveFile` without `WriteFile | Truncate` on Landlock, `file-write-create` / `file-write-unlink` with `file-write-data` still denied on Seatbelt.

That keeps the kernel rule as the control, works on Landlock-only hosts — where the reporter actually is — leaves both guard tests passing unchanged, and generalises to the next client that needs a lock, pid file or socket in a protected tree.

Being written now. This PR stays open until that lands, in case the overlay work is worth keeping for a different reason.

Starefossen:
Superseded by #455, which fixes this in the access model rather than as a per-agent exception.

Recorded here because the reasoning is worth keeping: this approach granted `~/.pi/agent` writable when bubblewrap is available and relied on the `ro_protect` overlay to re-bind `bin/` read-only. Three things were wrong with it.

It probably does not fix the reported case — the reporter got `EACCES`, which is Landlock's denial; a read-only bind mount gives `EROFS`, so bubblewrap was likely configured but not active, and a bwrap-gated fix does nothing for them. It made the overlay the sole control, where Landlock had been a second layer, and bubblewrap silently skips a re-bind whose source does not exist. And Linux CI caught that my first version dropped `bin/` from the grant list entirely, which would have removed it from the overlay too — `pi_managed_binaries_are_not_inside_a_writable_linux_grant` failed, correctly.

#455 adds a create-only directory grant instead: `mkdir`/`rmdir` with no `WriteFile`, so the kernel rule stays the control and both guard tests pass unchanged. It is under review for two escapes of its own — rename is a directory operation, so the grant needs scoping to the lock's name rather than the whole directory — but the shape is right and it does not trade a control away.

Issue #447, opened 2026-09-08
check exec reports a named repository as outside the gh startup scope


Starefossen:
`--repo-dir` / `sandbox.repo_dirs` exist so `gh` may target the named repository — the flag's help says exactly that. The launch implements it. `cplt check exec` did not.

Reproduced with two real repositories inside one umbrella repository, both named:

```
$ cplt check exec gh pr create -R navikt/bar
gh pr create -R navikt/bar: BLOCKED
  Reason: 'gh pr create' targets 'navikt/bar' which is outside the startup repo 'navikt/min-arbeidsflate'.

$ cplt exec -- gh pr create -R navikt/bar
(reaches gh — the guard allowed it)

$ cplt exec -- gh pr create -R navikt/other
⚠️ BLOCKED by sandbox: 'gh pr create' targets 'navikt/other' which is outside the
startup repo 'navikt/min-arbeidsflate, navikt/bar, navikt/baz'.
```

The launch's refusal names all three repositories in scope; `check exec`'s names one.

**Cause:** `explain_exec`'s gh arm called `gate(&rest, ctx.project_dir, &policy)`, which resolves scope from the project directory alone. The launch calls `gate_with_repo_scope` with the captured set. `run_check_command` already had the validated roots and discarded them — `let _ = &repo_roots;`, with a comment saying check does not yet report per named root.

Fixed in the PR that references this, along with a golden test asserting both halves: a named repository is allowed, and one that was never named is still refused.

**Worth recording as a pattern.** This is the third defect of the same shape in two days, after the two in #439 (`check exec` ignoring the enforcement mode, and ignoring that no scratch dir means no guard). Each time, `check exec` answered "what will happen" from fewer inputs than the launch has: no enforcement mode, no repository facts, no scratch-dir state, no scope set.

The structural fix is not another field. It is that `ExecContext` should be built from the same value the launch resolves from, so a new input cannot reach the launch without reaching this surface too — otherwise the fourth instance is already written. Worth folding into #427, which is rebuilding exactly that resolution path.

Starefossen:
Fixed in #448 (`78ce9d0`), and the structural half in #453 (`1f180f3`).

#448 built the gh scope set the way `sandbox_exec` builds it, so `check exec gh pr create -R <named repo>` now agrees with the launch, with a golden test asserting both halves — a named repository is allowed, one never named is still refused.

[...]

PR #454, opened 2026-09-08
feat(doctor): report which sandbox mechanism is actually enforcing

Starefossen:
#449 could not be diagnosed from anything cplt prints, and that is the actual finding.

The reporter said bubblewrap was "enabled" — meaning configured. Whether it was **active** decides what is enforced, and nothing reports that. I had to infer the regime from a Node errno in their traceback (`EACCES` is Landlock's denial; a read-only bind mount would have given `EROFS`), concluded bubblewrap probably was not running, and only then realised the fix I had proposed might not help the person who reported the bug.

On Linux the protections are split across two mechanisms, and several exist **only** under the mount overlay, because Landlock cannot deny a path inside a tree it has granted:

- `.git/hooks` and the rest of the git-persistence set
- the agent config files that auto-execute on the host next time it runs outside cplt
- mise's `shims/` and `installs/`
- Copilot's package dirs
- Pi's managed `fd`/`rg`

`doctor` reported the Landlock ABI and mentioned bubblewrap only inside a conditional caveat about escape sockets. So an operator on a host without it — WSL2 restricting user namespaces, a container, a minimal image — was told nothing about the protections they do not have.

It now says whether bubblewrap was found, and when it was not, names what is consequently unenforced and how to close it. **Availability, not configuration**: auto-detect falls back silently, and the fallback is precisely the case worth reporting.

## Agent grants

Second gap from the same report: nothing printed the agent's own grants.

```
[doctor] Agent grants (Copilot)
  ✓ read/write, execute  /Users/hans/.copilot
```

"The agent cannot write X" is unanswerable without them. For Pi the answer is one line — `~/.pi/agent` read-only with write carved back per subdirectory — and it was visible nowhere. Each entry shows access and whether it exists, because on Linux a path that does not exist gets no read-only re-bind, which turns a protection into a no-op silently.

Reporting only, no behaviour change. This is the prerequisite for asking the reporter anything useful, and for the next report of this shape being answerable from one command.

### DIFF
--- a/src/discover.rs
+++ b/src/discover.rs
@@ -1016,6 +1016,30 @@ fn print_sandbox_mechanism_status() -> bool {
             ui::stdout_color(ui::GREEN),
             ui::stdout_color(ui::RESET)
         );
+        // Whether bubblewrap is available decides what is actually enforced on
+        // this host, and nothing reported it. Several protections exist ONLY
+        // under bwrap — `.git/hooks`, the agent host-persistence files, mise's
+        // shims, Copilot's package dirs — because Landlock cannot subtract a
+        // deny from inside an allowed tree. Without this line an operator
+        // cannot tell which regime they are in, and neither can a bug report:
+        // #449 was diagnosed from a Node errno because this fact was missing.
+        if crate::sandbox::bwrap_available() {
+            println!(
+                "  {}✓{} Bubblewrap: available — mount-level protections active",
+                ui::stdout_color(ui::GREEN),
+                ui::stdout_color(ui::RESET)
+            );
+        } else {
+            println!(
+                "  {}⚠{} Bubblewrap: not found (looked on PATH)",
+                ui::stdout_color(ui::YELLOW),
+                ui::stdout_color(ui::RESET)
+            );
+            println!("      Landlock-only. It cannot deny a path inside a tree it has granted,");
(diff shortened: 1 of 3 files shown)


Starefossen:
Superseded by #456, which does this properly.

This PR added two facts to `doctor` — bubblewrap availability and the agent's grants — because #449 could not be diagnosed without them. Right instinct, wrong shape, and it carried a bug that made it useless for the case it was written for.

The bug: the "Agent grants" section resolved the agent with `Agent::auto_detect()`, which can only return Copilot, OpenCode or Antigravity — never Pi. It would have printed *Copilot's* grants to the #449 reporter. The same defect class as #447, in the fix for a diagnosis failure.

And two more sections on a 52-line dump was treating the symptom. #456 rebuilds `doctor` from `resolve_context`, so the agent and the enforcement regime are the ones a launch would actually use, prints computed findings instead of an inventory, and puts the inventory behind `--verbose`. The #449 diagnosis becomes one line rather than two facts the reader has to join.

The bubblewrap reporting survives there, and correctly three-valued — usable, installed-but-unusable, disabled by config — using the same functionality probe the launch runs. This PR's version reported "available" from a PATH lookup, which repeats the "configured is not active" trap that caused the misdiagnosis in the first place.

PR #455, opened 2026-09-08
feat(sandbox): add a create-only directory grant, and use it for Pi's trust lock

### LINKED ISSUE #449: Pi: Error: EACCES: permission denied, mkdir '/home/mark/.pi/agent/trust.json.lock' when running cplt on a repository I've already trusted

Starefossen:
Fixes #449 in the access model rather than as a per-agent exception, which is what the maintainer asked for: *"I can not imagine this is the last time we need this kind of exception… I want this fixed properly in our model… without hard coding in exceptions all over the place."*

Supersedes #452 — see the reasoning recorded there.

## The gap in the model

`AgentDir` already carries `write_files` for narrow carve-outs in a read-only tree, honoured by both backends. But it grants write on a path that **already exists**, and creating an entry is a right on the *parent* directory. So there was no way to express "may create its lock here", and the fix kept drifting toward per-agent, per-platform special cases.

`FsAccess::create_dirs` / `AgentDir::create_dirs` is the missing primitive: **a directory stays read-only for its existing contents while permitting creation and removal of new subdirectories.**

- **Landlock:** `MakeDir | RemoveDir`. No `WriteFile`, no `Truncate` — existing files stay unwritable.
- **Seatbelt:** `file-write-create` / `file-write-unlink`, both `(require-all (subpath …) (vnode-type DIRECTORY))`. `file-write-data` is never allowed, so `(deny default)` keeps existing files unwritable.

Declared on Pi's `~/.pi/agent`. The per-subdirectory writable grants and `write_files` are untouched; this is additive.

## It is narrower than I specified, for a good reason

I briefed `MakeDir | MakeReg | RemoveDir | RemoveFile`. The implementation is directory-only, and the argument is sound: on Seatbelt the regular-file variant **permits hardlink creation**, so `rm rg` followed by `ln <attacker-file> rg` would reconstitute the managed-binary name pointing at attacker bytes — an RCE, and precisely what the read-only `bin/` design exists to prevent.

Directory-only is also sufficient, verified rather than assumed: Pi 0.85.1 depends on `proper-lockfile` 4.1.2, which takes a lock by `mkdir`-ing `<file>.lock` — a directory, matching the reporter's `mkdir` error. I confirmed the dependency and the read path in the published npm tarball.

So the agent can only ever create an empty directory. `bin/` and the managed binaries stay fully immutable — a tighter bound than the brief asked for.

## The honest limit

Landlock cannot scope creation to a *name*. The bound is "existing files stay unwritable" and "only directories", not "only `trust.json.lock`". Stated in the field's doc comment.

The residual concern — creating a name that shadows an exec-only child — is closed because cplt pre-creates every agent dir before launching, so `bin/` is already taken, and a new directory cannot receive a file anyway. `create_dirs_exec_only_children_are_pre_created` runs that exact loop and asserts it, rather than leaving it implicit.

## Guard tests pass unchanged

`pi_config_dirs_keep_bin_out_of_every_writable_grant` and `pi_managed_binaries_are_not_inside_a_writable_linux_grant` assert `bin/` is never inside a writable grant. Both pass with **zero modifications** — verified: the diff against main deletes nothing from them. That was the requirement, because create-only is not `write`; if either had needed weakening the design would have been wrong. (Under the rejected #452 approach the second one failed.)

New tests drive the "stays read-only" half from the agent table, so a future agent adopting the capability is covered without anyone remembering.

## Bubblewrap

In scope in practice, though I had scoped it out: bwrap applies Landlock *inside* a `--ro-bind / /` namespace, so without a mount change `mkdir` fails with `EROFS` before Landlock is consulted and Pi still would not start. Create-only dirs join the writable-bind set.

This is not #452 wearing a different hat. There, Landlock granted the root **write** and the overlay was the sole control. Here Landlock grants create-only with no `WriteFile`, so `bin/` is write-denied by **both** Landlock and the overlay.

## Not exercised on macOS

Landlock and bubblewrap run only on Linux; the Seatbelt semantics were validated directly with `sandbox-exec`, and the Linux target cross-checks clean (`cargo check`/`clippy --target x86_64-unknown-linux-gnu`). Linux CI is the first place the ruleset actually runs.

Closes #449.


Starefossen:
Rewritten after an adversarial review found **two verified sandbox escapes** in the first version. Both had the same root cause: the grant was scoped to a *directory*, and rename is a directory operation.

## What was wrong

**Seatbelt.** The create-only allow overrode the tail host-persistence denies. Verified against the real generated Pi profile with `sandbox-exec`:

```
mv ~/.pi/agent/bin  bin.old     ALLOWED   (despite the deny on bin/)
mv ~/.pi/agent/tmp  bin         ALLOWED   → bin/rg is now attacker bytes
mv sessions extensions          ALLOWED   → planted extensions auto-load on the next host pi
```

Empirically `require-all(subpath, vnode-type)` **allow beats a plain `subpath` deny**; a `literal` deny wins. So the comment justifying the design ("denies are emitted later and win under last-match-wins") was false for that rule shape.

**Landlock.** Same-directory rename does not require `Refer` — `current_check_refer_path` in `security/landlock/fs.c` is explicit — so `MakeDir | RemoveDir` alone permits `mv bin bin.old; mv tmp bin`. The comment claiming "no Refer, no rename" protected nothing.

Both escapes passed every test in the first version, including the two guard tests I had made a hard requirement, because all of them reduce to `writable_tree_over`, which keys on the `write` flag this capability deliberately sidesteps.

## What it does now

**Scoped to a name.** `AgentDir::create_dirs` is a `Vec<&'static str>` (Pi: `["trust.json.lock"]`), emitted as one `(allow file-write* (require-all (literal "<dir>/<name>") (vnode-type DIRECTORY)))` per name — never `subpath`. `file-write*` rather than the individual ops because `proper-lockfile@4.1.2` probes mtime with `fs.utimes` immediately after `mkdir` and fails the lock on error; without it macOS would have moved from `EACCES mkdir` to `EPERM utime` and Pi still would not start.

**Withheld on Linux without bubblewrap.** There is no kernel-native way to stop that rename, and trading H-13 for Pi startup is not a default worth taking. `prepare_impl` clears the grant unless bubblewrap resolved a wrapper, and warns why. Under bwrap the mountpoints make the rename `EBUSY`. Where it is withheld, `cplt doctor`'s Pi finding now explains it — that is what #456 is for.

**Live probes, not string matching.** macOS asserts the real table under `sandbox-exec`; Linux equivalents through `cplt exec` for both regimes. I falsified the macOS one myself: reverting the emitter to the root-wide `subpath` form fails it with "mkdir of any other name must be denied".

Worth recording: the agent's own first probe run sat under `/private/tmp`, which the profile grants `file-write*` wholesale, so it passed vacuously. The live tests deliberately do not use a tempdir.

## Residual, stated rather than hidden

`mv tmp trust.json.lock` is still allowed — a populated directory at the lock's name. That breaks Pi's own lock and nothing else: self-DoS, no persistence, nothing the host later executes. Documented in the code and in known-impacts.

## Interaction with #456

Rebased onto it. `doctor::dir_is_read_only` keyed on `!access.write`, which a create-only grant sidesteps, so it would have reported "Pi will not start" on every working Pi host. Fixed here, since the field does not exist without this PR. That is the generalised cost of a new access dimension, and the reason to look for others.

Linux CI is the first place the Landlock and bubblewrap paths actually run.

PR #487, opened 2026-09-10
feat(git-guard): say when branch tracking is about to be silently dropped

### LINKED ISSUE #402: Branch tracking silently dropped: .git/config deny exits 0 and reports success
## Summary

Commands that record branch tracking hit the `.git/config` write deny, print the error, and then **exit 0 while claiming success**. The tracking config is silently dropped.

```
error: could not write config file .git/config: Operation not permitted
branch 'feat' set up to track 'origin/main'.
```

The second line is false, and the exit code is 0, so an agent checking exit status sees a clean run.

## Measured behaviour

macOS, scratch repo with a real bare origin, under `cplt exec`:

| command | result |
| --- | --- |
| `git switch -c feat2` (no tracking) | clean, exit 0 |
| `git checkout -b feat origin/main` | branch created, exit 0, deny on stderr, false "set up to track" line |
| `git branch --set-upstream-to=origin/main feat2` | same |
| `git push -u origin feat3` | push succeeds, upstream not recorded, exit 0, same false line |
| `git config user.name bob` | exit 4 |

## Why it happens

`emit_git_persistence_denies` / `emit_gitdir_denies` (src/sandbox_profile.rs:860-886) emit `(deny file-write* (literal "<gitdir>/config"))`, asserted at src/sandbox_profile.rs:2231, 2918, 2984. The rule is deliberate — `url.*.insteadOf` hijacking and `core.hooksPath` (docs/security.md:94) — and macOS-only.

The deny is documented (docs/known-impacts.md:351, :353, :356; docs/git-guard.md:225, 240). What is not documented is which everyday commands hit it. docs/known-impacts.md:347 says `git checkout/merge/rebase/branch` "✅ Works — Branch operations work normally", which is not true for the tracking variants.

## Suggested fix

Amend the `docs/known-impacts.md:347` row and add a line to the git table: `git checkout -b <remote>/<branch>`, `git branch --set-upstream-to` and `git push -u` succeed but do not save upstream tracking on macOS; set it outside the sandbox, or use explicit `git push origin HEAD`.

Suppressing the misleading "set up to track" line is not feasible — git prints it itself. Documenting it is the whole fix.

Starefossen:
The behaviour half of #402, per the decision to warn from the wrapper rather than intercept. The documentation half landed in #480.

## The problem

`.git/config` is write-denied on macOS. git treats the failed write as non-fatal:

```
error: could not write config file .git/config: Operation not permitted
branch 'feat' set up to track 'origin/main'.
```

The second line is false and the exit code is 0. An agent reading the summary line, or checking the exit status, is told the tracking was recorded. A field agent today read stderr and caught it — correctly, and said so. The next one may not.

## The change

A `push`, `branch`, `checkout` or `switch` carrying `-u`, `--set-upstream`, `--set-upstream-to` or `--track` gets a note before it runs:

```
cplt: this records branch tracking in .git/config, which the sandbox denies. git will print
`set up to track` and exit 0 anyway — that line is false and the upstream is NOT recorded.
Push with an explicit refspec and name the branch when you need it:
`git push origin HEAD:my-branch`, then `gh pr create -R <owner>/<repo> --head my-branch`.
The same applies to `git checkout -b <name> <remote>/<base>`.
```

Not intercepting, per the decision: the push itself succeeds and refusing it would break something that works.

## Three deliberate limits

- **macOS only.** Landlock cannot deny a file inside a writable directory, so on Linux `.git/config` stays writable and tracking is recorded normally. Warning there would be the false statement.
- **Explicit flags only.** `git checkout -b <name> <remote>/<base>` also sets tracking, from a positional only git can distinguish from a pathspec. The note names that form rather than this guessing at it.
- **Only where nothing else is being said.** A refusal, or a warn-mode notice, is the more urgent message; two notices on one command is how both get skimmed.

It also needs the git guard on, since the wrapper it prints from is the guard's. Stated in `known-impacts.md` rather than left to be discovered — that is the same gap #478 fixed for `inject_token`, and it is worth naming before someone hits it.

## Verification

`tracking_flags_are_noticed_and_ordinary_commands_are_not` covers five tracking forms and five that must stay quiet, including `git clean -u`, where `-u` is a different flag entirely. Falsified by removing the subcommand scope — it fails.

Verified in a real session: `git push -u origin feat` prints the note and the push proceeds.

One note on getting there: my first attempt printed nothing, and I nearly reported it as broken. The push was being refused because the fixture had no recorded `refs/remotes/origin/HEAD`, so the guard failed closed before reaching the notice — working as designed, bad fixture.

Closes #402's behaviour half.

### DIFF
--- a/docs/known-impacts.md
+++ b/docs/known-impacts.md
@@ -29,6 +29,14 @@ tracking was recorded. It was not.
 | `git config user.name <x>` | Fails loudly, exit 4 |
 | `git switch -c <branch>` (no tracking) | Clean, nothing to record |
 
+cplt says so when it can see the request coming: a `push`, `branch`, `checkout`
+or `switch` carrying `-u`, `--set-upstream`, `--set-upstream-to` or `--track`
+gets a note before the command runs, saying the success line will be false. That
+needs the git guard on, since the wrapper it prints from is the guard's. The
+implicit form, `git checkout -b <name> <remote>/<base>`, is named in the note
+rather than detected: it sets tracking from a positional that only git can tell
+from a pathspec.
+
 **The shape that works**, and what to tell an agent to do: push with an explicit
 refspec and open the PR with an explicit head, so nothing needs local config.
 
--- a/src/gh_proxy.rs
+++ b/src/gh_proxy.rs
@@ -2709,21 +2709,8 @@ pub fn gate_git(
     }
 
     // Find the subcommand by skipping global flags.
-    let mut i = 0;
-    let mut subcommand = None;
-    while i < args.len() {
-        let arg = args[i];
-        if GIT_GLOBAL_FLAGS_WITH_VALUE.contains(&arg) {
-            i += 2; // skip flag and its value
-            continue;
-        }
-        if arg.starts_with('-') {
-            i += 1;
-            continue;
-        }
-        subcommand = Some(arg);
-        break;
-    }
+    let i = git_subcommand_index(args).unwrap_or(args.len());
+    let subcommand = args.get(i).copied();
 
     // Defense in depth: block command-line config (`-c` / `--config-env`) that
     // could redirect where a push lands or redefine what a subcommand does.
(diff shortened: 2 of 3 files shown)


Starefossen:
Good catch — fixed in the last commit.

`tracking_flag` took the first token not starting with `-`, so `git -C /elsewhere push -u origin feat` read `/elsewhere` as the subcommand and printed nothing. That is exactly the case the notice exists for: an agent pushing from a linked repository.

The push guard next door already had this scan, skipping global flags and the values they consume. I pulled it out as `git_subcommand_index` and both call it, so the two cannot drift apart on where a git command starts. The tracking-flag scan now also starts after the subcommand, since a global flag's value is not a flag.

Test covers `-C <path>` and `-c <key=value>` before `push`.

PR #489, opened 2026-09-10
fix(check): run the probes under /bin/sh, not the user's login shell

### LINKED ISSUE #488: cplt check reports a healthy sandbox as NOT ENFORCING for fish users
## What happens

On a host whose `$SHELL` is fish, `cplt check` runs its probes through fish, which cannot parse them:

```
fish: Unsupported use of '='. In fish, please use 'set f '/Users/hans/go/src/github.com/nais/unleash'/.cplt-check-write.$$'.
exec >/dev/null 2>&1; f='/Users/hans/go/src/github.com/nais/unleash'/.cplt-check-write.$$; : > "$f" && rm -f "$f"
```

The parse error exits non-zero, `decode_fs_probe` reads any non-zero as denied, and every write probe reports BLOCKED:

```
  ✗ write project dir          → BLOCKED  [expected ALLOWED]
  ✗ write named repo bifrost   → BLOCKED  [expected ALLOWED]
→ Sandbox is NOT ENFORCING. 3 expected-allowed check(s) were blocked
```

The sandbox is fine. The probe is what broke. `nav-pilot doctor` repeats the verdict, so its Project Security section also says the sandbox is not enforcing.

Worse than a cosmetic wrong answer: `check` is the command we point people at when something looks off, and it tells a fish user their sandbox is broken in a way they cannot act on. Every fish user on macOS sees this, and fish is not rare among the people running cplt.

## Cause

`probe_shell` resolved its interpreter with `Agent::Shell.resolve_binary()`, which is `$SHELL`. That is right for `cplt --agent shell` and for `cplt -c` — the user asked for their shell there. The probes are not the user's shell: they are POSIX scripts cplt writes itself, using `f=...` assignments, `$$`, and `exec >/dev/null 2>&1`.

The read probe passing while the write probe failed is what made it look like a sandbox verdict rather than a parse failure — `: < path` happens to survive fish.

## Fix

Run the probes under `/bin/sh`. The `-c` path keeps `$SHELL`, since that one is the user's request.

Starefossen:
Closes #488. Reported from a real session: `cplt check` in `nais/unleash` said

```
  ✗ write project dir          → BLOCKED  [expected ALLOWED]
→ Sandbox is NOT ENFORCING. 3 expected-allowed check(s) were blocked
```

on a sandbox that was enforcing perfectly.

## Cause

`probe_shell` resolved its interpreter with `Agent::Shell.resolve_binary()` — that is `$SHELL`. The probes are POSIX scripts cplt writes itself (`f=...`, `$$`, `exec >/dev/null 2>&1`), and fish rejects the assignment:

```
fish: Unsupported use of '='. In fish, please use 'set f ...'
```

Non-zero exit, and `decode_fs_probe` reads any non-zero as denied. The read probe passed because `: < path` happens to survive fish, so the output looked like a sandbox verdict rather than a parse failure.

## Fix

One line: `/bin/sh`. The `cplt -c` path keeps `$SHELL` — there the user did ask for their own shell, and the command is theirs, not ours.

## Why this one matters more than its size

`check` is the command we send people to when something looks wrong. Telling a fish user their sandbox is not enforcing, with no actionable cause, is the failure mode that makes someone turn cplt off. `nav-pilot doctor` repeats the verdict, so it was saying the same thing.

## Verification

Before and after on the reporting host, with `SHELL=/opt/homebrew/bin/fish`:

```
✗ write project dir → BLOCKED  ...  Sandbox is NOT ENFORCING
✓ write project dir → ALLOWED  ...  Sandbox is ENFORCING (6 protections verified)
```

`e2e_check_probes_ignore_a_hostile_user_shell` points `$SHELL` at a stub that fails whatever it is handed — same shape as fish refusing a POSIX assignment, no fish needed on the runner. Falsified by restoring the old resolution: it fails with NOT ENFORCING.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -4768,6 +4768,12 @@ fn probe_shell(
     // is not an assignment there, so the write probe exits non-zero and check
     // reports the sandbox as not enforcing (#488).
     let shell = PathBuf::from("/bin/sh");
+    // A missing `/bin/sh` is a probe that could not run, not a denial. Without
+    // this the spawn failure decodes as BLOCKED and `check` reports the sandbox
+    // as not enforcing for a reason that has nothing to do with the sandbox.
+    if !shell.exists() {
+        return PROBE_SPAWN_FAILED;
+    }
     let args = vec!["-c".to_string(), script.to_string()];
     sandbox::exec_sandboxed(
         prepared,
--- a/src/main.rs
+++ b/src/main.rs
@@ -4763,9 +4763,11 @@ fn probe_shell(
     disabled: &[sandbox::HardeningCategory],
     script: &str,
 ) -> u8 {
-    let Ok(shell) = agent::Agent::Shell.resolve_binary() else {
-        return PROBE_SPAWN_FAILED;
-    };
+    // `/bin/sh`, never `$SHELL`: the probes are POSIX shell, and a user whose
+    // login shell is fish gets a syntax error instead of a verdict — `f=...`
+    // is not an assignment there, so the write probe exits non-zero and check
+    // reports the sandbox as not enforcing (#488).
+    let shell = PathBuf::from("/bin/sh");
     let args = vec!["-c".to_string(), script.to_string()];
     sandbox::exec_sandboxed(
         prepared,
--- a/tests/e2e.rs
+++ b/tests/e2e.rs
@@ -5786,6 +5786,10 @@ paths = [
 
         let stdout = String::from_utf8_lossy(&output.stdout);
         assert!(
+            output.status.success(),
+            "an enforcing battery exits 0:\n{stdout}"
+        );
+        assert!(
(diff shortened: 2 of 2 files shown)


Starefossen:
Both fixed.

**`PROBE_SPAWN_FAILED` unreachable** — correct, and it matters: with the interpreter no longer resolved through `Agent::Shell`, nothing returned that code, so a host without `/bin/sh` would have had its spawn failure decoded as BLOCKED. That is the same wrong verdict this PR exists to remove, arriving by a different route. `probe_shell` now returns `PROBE_SPAWN_FAILED` when `/bin/sh` is not there, and the Inconclusive branch is live again.

**Test not asserting exit status** — added. The battery exits 0 when enforcing, so asserting it costs nothing and closes the gap where the command errors but still prints the string.

PR #490, opened 2026-09-10
feat(config): resolve relative paths when setting them, and add -g/-l/-r

Starefossen:
Two small things a maintainer hit while linking repositories by hand.

## `../sibling` now works

```
$ cplt config set --local sandbox.repo_dirs ../sykepenger-model
```

Before, this was refused with "local config accepts absolute and ~/ paths only" and the user got to work out the absolute path themselves. cplt knows the working directory.

The rule the refusal was protecting is still intact, and is the reason the *resolved* path is what gets stored: a relative entry inside the file has no stable anchor, so it would name different directories depending on where cplt was started. What changes is where the resolution happens — at write time, once, instead of never.

Details worth stating:

- `~/` is stored as typed. It is already anchor-independent, and expanding it would bake one machine's `$HOME` into a file that may be read on another.
- The path must exist when set. A relative entry cplt cannot resolve now is one the next launch would refuse anyway; failing at write time names the directory you were standing in, which is the fact you need.
- `--unset` resolves the same way, so a root can be removed by the same name it was added under.
- Canonicalized, so the entry survives a later `cd` and names the tree the user actually pointed at.

`is_local_path_key` reads the same table as the load-time check, with a test asserting the two agree — otherwise a future path key gets resolved but not validated, or validated but not resolved.

## `-g` / `-l` / `-r`

Short forms for `--global`, `--local` and `--repo` on `config set`, `-l` on `config path`, `-g` on `init`. These are the flags typed most often. Docs keep the long forms, since they name the file being written.

## Verification

`cargo test` clean, clippy clean, `--target x86_64-unknown-linux-gnu --all-targets` clean. Four new unit tests: resolution, `~/` and absolute passthrough, the nonexistent-path refusal, and the agreement between the resolve list and the validate list. End-to-end: set with `../../nais/unleash` from this repo, `config show` reports the absolute path, `--unset` with the same relative string removes it.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -209,6 +209,34 @@ it.
 Each refusal names the layer that does work, so you do not have to remember the
 table.
 
+`config set` takes `-g`, `-l` and `-r` for `--global`, `--local` and `--repo`.
+The long forms are used throughout this document because they say which file is
+being written.
+
+### Relative paths are resolved when you set them
+
+`../sibling` is how people name the repository next door, so `config set` takes
+it and stores what it resolves to:
+
+```bash
+cd ~/src/spleis
+cplt config set --local sandbox.repo_dirs ../sykepenger-model
+# [cplt] sandbox.repo_dirs = [/Users/you/src/sykepenger-model]
+```
+
+The file holds absolute paths only. A relative entry *in* the file has no
+stable anchor — it would mean different directories depending on where cplt was
+started — so what gets written is the resolved path, not what you typed. A
+`~/` entry is stored as typed: it is already anchor-independent, and expanding
+it would bake in one machine's home directory.
+
+The path has to exist when you set it. A relative entry cplt cannot resolve now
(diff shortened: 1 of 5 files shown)


Starefossen:
Fixed, and it was a real hole rather than a tidiness point.

`~someone/repo` passed the absolute-or-tilde check, but `expand_tilde` (`src/config/path.rs:324`) only handles `~` and `~/…` — so the entry was stored as if anchor-independent and would then reach `validate_repo_dirs` as a relative path resolved against whatever directory cplt was started in. That is precisely what the relative-path refusal exists to prevent.

Both places now use the same rule: `/`, `~`, or `~/…`. Refused rather than resolved, since guessing which user's home was meant is worse than saying so. The load-time check got the same tightening, so an existing file carrying such an entry is refused instead of quietly reinterpreted.

Test covers `~someone/repo` and `~root` at both write and load.

Issue #491, opened 2026-09-10
Design: propose linked repositories by identity, resolve and verify them locally


Starefossen:
Design for the feature #446 recorded as permitted but not yet worth building. #344 has since landed, including siblings, so the sequencing objection in #446 ("it would mostly resolve repositories it then has to refuse") no longer holds. What follows is a concrete proposal, grounded in the machinery that exists today.

The friction that prompted it: a team knows their sessions span three repositories, and every developer has to discover that fact and then run `cplt config set --local sandbox.repo_dirs <path>` three times with paths only they know. The project cannot say what the project is.

## What a repository may say, and what it may not

```toml
[propose]
repos = ["navikt/sykepenger-model", "navikt/spleis-testdata"]
```

An identity, never a path. `<owner>/<name>`, matched against the origin of a candidate checkout — the same normalisation `trust::normalize_remote_url` already does, and the same forward map `gh_proxy::detect_current_repo` already uses for the launch summary.

This is the distinction the layer split already rests on, and the reason the current refusal does not forbid it:

> naming other trees as project-grade roots is a path grant, and repo config cannot grant paths

`navikt/sykepenger-model` grants nothing. It names a repository. The path is chosen on the user's machine, by the user, and stored in the local layer exactly as it is today. `.cplt.toml` gains no ability to point at a directory of its choosing.

## Finding it

cplt looks in this order and stops at the first verified hit. Every candidate must clear the identity check below before it counts as found.

1. **Already linked.** The path is in `sandbox.repo_dirs` for this checkout. Nothing to do; the proposal is satisfied.
2. **Siblings.** `<parent>/<name>`, where parent is the launch repository's parent directory. This is the overwhelmingly common layout and it is a single `stat`.
3. **Same forge, other org.** When the launch path ends `<host>/<org>/<repo>` — the Go layout, `~/go/src/github.com/navikt/cplt`, and anyone who mirrors it — also try `<host>/<proposed-org>/<proposed-name>`. This is what makes `nais/unleash` findable from `navikt/cplt`, and it is one more `stat`.
4. **Named search roots.** `sandbox.repo_search_paths` in the local layer, for people whose checkouts are not where the first three look. Depth 2, so `~/src/<org>/<repo>` and `~/src/<repo>` both resolve.
[...]

Issue #443, opened 2026-09-08
The pass_env repo refusal claims environment variables are machine-specific, which is often false


Starefossen:
`cplt config set --repo sandbox.pass_env FOO` is refused with:

```
[cplt] sandbox.pass_env is not valid in repo config.
  Reason: environment variables are machine-specific, not project policy.
```

The premise is too strong. Plenty of environment variables a project needs are properties of the **application**, not of the machine it happens to be built on, and plenty are not sensitive at all: `NODE_ENV`, `TZ`, `LANG`, `SPRING_PROFILES_ACTIVE`, `NAIS_CLUSTER_NAME`, a feature-flag name a test suite reads. If a project's tests need `TZ=Europe/Oslo` to pass, that is project policy in exactly the sense `allow.ports 5432` is, and today every developer has to discover it and set it themselves.

So the refusal is doing something defensible for a reason it states wrongly, which makes it hard to argue with and hard to learn from.

## The real reason, as far as I can tell

The hazard is not that env vars are machine-specific. It is that `pass_env` names variables in the **parent's** environment, so a repository that can set it is asking to pull whatever the host happens to have under that name into the agent's environment. A committed `pass_env = ["GITHUB_TOKEN"]` — or `AWS_SECRET_ACCESS_KEY`, or something whose name collides with a variable the developer sets for unrelated reasons — is a credential exfiltration request that reads as ordinary configuration in a diff.

That is a real reason to keep it out of `[deny]`-style automatic application. It is not obviously a reason to keep it out of `[propose]`, which exists precisely for "the repo asks, a human accepts on this machine, and an edit to the file de-activates the approval".

## Options

1. **Fix the message only.** Keep the refusal, say what it is actually protecting against: a repo naming a variable pulls whatever the host has under that name, so it is not something a committed file may decide. Cheapest, and stops the message losing arguments it should win.
2. **Allow it under `[propose]`.** Then a project can state the variables its build genuinely needs, the reviewer sees the list in the PR, and each developer accepts once. The trust store's content hash already handles the "someone added `GITHUB_TOKEN` later" case — the approval de-activates.
3. **Allow it under `[propose]` with a name filter.** As above, but refuse names matching the credential-ish patterns cplt already knows about (`*_TOKEN`, `*_SECRET`, `*_KEY`, `*_PASSWORD`, `AWS_*`), so the common accident is refused at write time and the deliberate case still needs a human. A denylist here is a heuristic, not a boundary, and would need to be described as such.

[...]

PR #493, opened 2026-09-10
fix(trust): hash the whole proposal, so an approval cannot outlive its values

### LINKED ISSUE #492: An approval survives a change to pass_env: the proposal hash never covered it
`[propose] pass_env` has never been part of `proposal_content_hash`, so the content pinning that is supposed to invalidate an approval when the proposal changes does not apply to it. `gradle_init`, `gh_guard`, `git_push_prevention` and the collected unknown keys are missing the same way.

Reproduced end to end on `main` (`a91f636`):

```
$ printf '[propose]\npass_env = ["TZ"]\n' > .cplt.toml && git commit -qam x
$ cplt trust accept --all
✓ Approved 1 permission(s): sandbox.pass_env

$ printf '[propose]\npass_env = ["AWS_SECRET_ACCESS_KEY"]\n' > .cplt.toml && git commit -qam y
$ AWS_SECRET_ACCESS_KEY=leaked-value cplt --agent shell -- -c 'echo GOT:$AWS_SECRET_ACCESS_KEY'
GOT:leaked-value
```

No re-prompt, no staleness warning. The trust file still lists `sandbox.pass_env` as approved and its hash still matches, because the hash was computed over a list of fields that does not include `pass_env`.

## Why it matters

`pass_env` names a variable in the *parent's* environment, so the value is whatever the developer's machine has under that name — the tokens and cloud credentials in a normal shell. The whole reason it is a proposal rather than a plain config key is that a reviewer sees the list in the diff and each developer approves it on their own machine. The doc comment on the field says so in as many words:

> each developer accepts it on their own machine — where an edit changes the file's hash and de-activates the approval (#443)

That sentence was not true when it was written. A repository approved once for a harmless `TZ` can later commit any variable name and collect it silently from every developer who approved the first version.

Both halves of the delivery chain have to be there for the guarantee to hold, and only one was.

## Cause

`proposal_content_hash` (`src/trust.rs:298`) is a hand-maintained list of fields to hash. `pass_env` was added to `ProposeSection` by #443 and never added there. It is the second time the list has fallen behind the struct — `allow.domains` needed the same manual addition in #486, and got it only because that PR touched the hash for other reasons.

A list that must be updated in a second file every time a proposal is added will go out of date again.

## Fix

Serialize the whole `[propose]` section and hash that, so a new field is covered by construction. Found by an adversarial review of #491, which was leaning on "covered by the content hash" as existing behaviour.

Starefossen:
Closes #492. Found by an adversarial design review of #491, which was leaning on "an edit de-activates the approval" as existing machinery. For `pass_env` it was not.

## The bug

Reproduced on `main` before the fix:

```
$ printf '[propose]\npass_env = ["TZ"]\n' > .cplt.toml && git commit -qam x
$ cplt trust accept --all          #  ✓ Approved: sandbox.pass_env
$ printf '[propose]\npass_env = ["AWS_SECRET_ACCESS_KEY"]\n' > .cplt.toml && git commit -qam y
$ AWS_SECRET_ACCESS_KEY=leaked-value cplt --agent shell -- -c 'echo GOT:$AWS_SECRET_ACCESS_KEY'
GOT:leaked-value
```

And after:

```
[cplt] .cplt.toml permissions changed since the last approval. Re-approve with `cplt trust accept`
GOT:
```

`pass_env` names a variable in the *parent's* environment, so the value is whatever the developer's machine holds — which is why it is a proposal a human approves rather than a config key. The field's doc comment claims the guarantee this PR implements; it has been wrong since #443.

Also unpinned, the same way: `gradle_init`, `gh_guard`, `git_push_prevention`, and the unknown-key map.

## The fix

`proposal_content_hash` was a hand-maintained list of fields. #486 had to add `allow.domains` to it by hand; #443 forgot. A list kept in a second file from the struct it mirrors will fall behind again, so the section is now serialized whole and the serialization is what gets hashed.

Two deliberate details:

- **JSON, not TOML.** `unknown` is `#[serde(flatten)]`, and TOML refuses a value emitted after a table — a future unknown scalar key would have made the hash fail to serialize and fall back to the `Debug` string. JSON has no such ordering rule.
- **Lists are sorted before hashing**, preserving the existing property that reordering a proposal is not a change. A future list field that nobody sorts hashes in written order: stricter, never looser.

`unknown` is hashed deliberately. A key this version cannot see grants nothing, but the next version may understand it, and an approval given while it was inert must not survive into a cplt that acts on it.

## Migration

Existing approvals go stale once and need one `cplt trust accept`. That is the fail-closed direction and exactly what the staleness path is for — the same one-time re-approval `approval_is_stale` already forces on trust files written before content pinning existed.

## Tests

`every_proposed_field_is_pinned` destructures `ProposeSection` exhaustively, so **adding a field to `[propose]` breaks this test to compile** — the moment to check the new field is covered. It then mutates each of the 21 proposable values one at a time and asserts every one changes the hash, and that no two changes collide. Falsified against the old implementation: `pass_env`, `gradle_init`, `gh_guard`, `git_push_prevention` and `unknown` all fail.

`a_changed_pass_env_value_invalidates_the_approval` pins the reported shape directly: approved for `["TZ"]`, later `["AWS_SECRET_ACCESS_KEY"]`, must read as stale.

The existing order-independence test still passes, which is why the sort is there.

### DIFF
--- a/src/repo_config.rs
+++ b/src/repo_config.rs
@@ -8,7 +8,7 @@
 //! The file is read from `git HEAD` (committed state) to prevent the sandboxed
 //! agent from modifying its own config mid-session.
 
-use serde::Deserialize;
+use serde::{Deserialize, Serialize};
 use std::path::{Path, PathBuf};
 
 /// The filename we look for in the project root.
@@ -67,7 +67,7 @@ pub struct DenySection {
 /// trusted on its own. A stray `preset` key lands in [`Self::unknown`] and is
 /// ignored, which is the same outcome `deny_unknown_fields` used to produce by
 /// refusing the file — without taking the rest of the config down with it.
-#[derive(Clone, Debug, Default, Deserialize, PartialEq)]
+#[derive(Clone, Debug, Default, Deserialize, Serialize, PartialEq)]
 #[serde(default)]
 pub struct ProposeSection {
     pub allow_localhost_any: Option<bool>,
@@ -116,7 +116,7 @@ pub struct ProposeSection {
 }
 
 /// Proposed allow expansions (paths, ports).
-#[derive(Clone, Debug, Default, Deserialize, PartialEq)]
+#[derive(Clone, Debug, Default, Deserialize, Serialize, PartialEq)]
 #[serde(default)]
 pub struct ProposeAllowSection {
     #[serde(default)]
@@ -144,7 +144,7 @@ pub struct ProposeAllowSection {
 }
 
 /// Proposed proxy settings.
-#[derive(Clone, Debug, Default, Deserialize, PartialEq)]
+#[derive(Clone, Debug, Default, Deserialize, Serialize, PartialEq)]
 #[serde(default)]
 pub struct ProposeProxySection {
     #[serde(default)]
--- a/src/trust.rs
+++ b/src/trust.rs
@@ -294,105 +294,51 @@ pub fn now_iso8601() -> String {
 ///
 /// Arrays are sorted before hashing so reordering entries does not
 /// invalidate approvals. Full SHA-256 hex is used (64 chars) for
-/// collision resistance.
+/// Hash of everything a repository proposes, pinning an approval to the exact
+/// values the user reviewed.
+///
+/// A stored hash that no longer matches makes the approval stale
+/// ([`approval_is_stale`]), so the keys stop applying until the user approves
+/// again. That is the whole defence against a repository being approved for one
+/// thing and later committing another.
+///
 pub fn proposal_content_hash(propose: &crate::repo_config::ProposeSection) -> String {
     use sha2::{Digest, Sha256};
 
+    // Serialized whole, never field by field. The hand-written version listed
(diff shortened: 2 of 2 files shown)


Starefossen:
Both fixed.

**Silent Debug fallback** — right that this is the wrong shape for a security boundary. The fallback is now tagged and carries the error (`cplt-proposal-hash-fallback:<err>:<debug>`), so it can never collide with a real serialization and never switches format quietly. It stays unreachable for plain data, which is why it is a fallback rather than a panic.

**Inconsistent doc block** — my edit left the old argument for *not* hashing unknown keys standing, plus a dangling half-sentence. Rewritten to state the two behaviours that do not read off the code: arrays are sorted first, and unknown keys are hashed.

On that second one, the doc now also says why it is worth its cost. Hashing unknown keys means an inert addition costs one re-approval. The alternative is a hash that deliberately ignores part of the file it exists to pin — and it was exactly that shape of exception that let an approval for `pass_env = ["TZ"]` survive into `pass_env = ["AWS_SECRET_ACCESS_KEY"]`.

PR #494, opened 2026-09-10
fix(repo-config): an uncommitted .cplt.toml applies its deny and proposes nothing

### LINKED ISSUE #206: Decide the repo-config trust boundary: macOS/Linux divergence and the --accept-repo-config bypass
Two questions surfaced while fixing #183 (PR #201). Neither is a bug to fix — both need a decision about what cplt's repo-config trust model should be — and they share a root cause, which is why they are filed together.

## 1. The same repo grants different permissions depending on the OS

`src/repo_config.rs:136-153`: on Linux the working-tree fallback is skipped, because Landlock cannot deny individual file writes inside the project directory — a sandboxed agent could rewrite `.cplt.toml` mid-session. On macOS the working-tree file is loaded and applied, with a `⚠ not committed` warning.

Deliberate since #34 (`f6969ab`). The consequence is that an untracked `.cplt.toml` is **live on macOS and inert on Linux**. Same repo, same command, different sandbox.

That matters beyond tidiness:

- A developer on macOS can be working under permissions that CI or a Linux colleague does not have, and neither side gets told.
- The diagnostics added in #201 had to be made `cfg!(target_os = "linux")`-aware precisely because a single honest sentence describing both platforms does not exist. That is a smell: the user-facing explanation is forked because the behaviour is.
- The tamper-resistance argument (read from git HEAD so an agent cannot grant itself permissions mid-session) is the stated security property, and on macOS it does not hold for the working-tree path.

Neither direction is obviously right:

- **Remove the macOS fallback.** Consistent, and the security property holds everywhere. Breaks every existing user who relies on an uncommitted `.cplt.toml` while iterating on their config, and there is no in-band migration.
- **Add a Linux fallback.** Consistent the other way, but weakens tamper-resistance on the platform where Landlock cannot compensate — which is exactly why it was skipped.
- **Keep the split, document it loudly.** Cheapest, and #201's messages already explain each platform correctly. But the divergence stays, and it stays surprising.

## 2. `--accept-repo-config` bypasses the commit guard entirely

`src/main.rs:1367` approves every proposed key with no state check at all. It is explicit per-run and does not persist to the trust store, so it is not the same severity as a persisted approval — but it means an uncommitted, gitignored, or non-git `.cplt.toml` is fully honoured for that run.

## Why these belong together

The commit guard lives in `trust_accept`, not at the `load_repo_config` boundary. That is the structural reason sibling cases keep having to be found one at a time:

- #183/#201 found that `git status --porcelain` is silent for a gitignored file, so a gitignored `.cplt.toml` sailed past the guard and was approved on macOS.
- The same PR then found `NotAGitRepo { has_file: true }` was still approvable, in a *different* branch of the same guard.
- `--accept-repo-config` is the third, and there is no reason to believe it is the last.

Moving the guard to the boundary would fix all three shapes at once and make a fourth structurally impossible. But it is a policy change, not a refactor: it would make an uncommitted `.cplt.toml` inert on macOS — i.e. it decides question 1 in the "remove the fallback" direction as a side effect.

That coupling is the actual decision to make. Fixing 2 properly settles 1.

## Not urgent

#201 closed the exploitable paths that persist to the trust store. What remains is a consistency and a bypass question, both known and now written down.

Starefossen:
Closes #206. Both questions in that issue, because — as it says — fixing the second settles the first.

## What changes

| | Before (macOS) | Before (Linux) | After (both) |
| --- | --- | --- | --- |
| Committed `.cplt.toml` | `[deny]` + approvable `[propose]` | same | unchanged |
| Uncommitted `.cplt.toml` | `[deny]` applies, `[propose]` unapprovable via `trust accept` **but granted by `--accept-repo-config`** | ignored entirely | `[deny]` applies, `[propose]` ignored |

## Why this way

**The fork was the code admitting it could not state itself honestly.** The working-tree fallback was skipped on Linux because Landlock cannot deny a write to a single file inside the project directory, so an agent could rewrite `.cplt.toml` between sessions. Fair — but that argues for not trusting *grants* from an unaudited file, not for discarding a repository's *restrictions* on one platform. The tell was `UNCOMMITTED_EFFECT`, a user-facing sentence written as `if cfg!(target_os = "linux")`: when the explanation has to fork, the behaviour has.

**A tightening and a relaxation are not the same operation.** `[deny]` can only reduce what the session may do. Dropping it costs a restriction its author believed was in force, and nothing else in the session will ever mention it. `[propose]` widens the sandbox, so it needs an audit trail, and a commit is what makes one.

**The guard belonged at the boundary, not at one door.** It lived in `trust_accept`, which meant every other route had to remember the rule. Three did not, and were found one at a time: a gitignored file (#183 — `git status --porcelain` is silent for one), a non-git directory (same PR, different branch of the same guard), and `--accept-repo-config`. Stripping `[propose]` in `load_repo_config` makes a fourth structurally impossible: there is nothing left downstream to approve.

`--accept-repo-config` therefore needed no code change. It still skips the prompt and still persists nothing — it just cannot approve what a commit never contained. CI checks out committed code, so the legitimate use loses nothing.

## What a user sees

```
[cplt] Repo config: .cplt.toml (working tree)
[cplt] The [propose] section of this uncommitted .cplt.toml is ignored. Commit the
       file to propose permissions; its [deny] keys apply either way.
```

and in `cplt trust`, where "No additional permissions requested" would have been a lie:

```
[cplt]  The [propose] section is ignored while the file is uncommitted.
```

`trust accept` now checks the commit state *before* asking whether anything is proposed — the other order answers "nothing to approve" for a file full of proposals and never mentions the commit.

## Verification

Three loader tests: an uncommitted file keeps its deny and loses its propose; the same file committed proposes normally; a deny-only uncommitted file reports nothing dropped (a notice that fires when nothing happened is how the real one stops being read). Falsified by removing the strip — the first fails.

`e2e_accept_repo_config_cannot_approve_an_uncommitted_proposal` runs the flag against an untracked config proposing `allow_docker` and asserts Docker is not granted and the reason is printed.

Manually, on a scratch repo: an uncommitted `[deny] paths = ["secrets"]` still blocks `cat secrets/s` (exit 1), `--accept-repo-config` grants nothing, and committing the file makes `trust accept --all` work as before.

`cargo test`, clippy, and `--target x86_64-unknown-linux-gnu --all-targets` all clean.

## Behaviour change worth calling out

A macOS user who iterates on `[propose]` in an uncommitted `.cplt.toml` loses that: those keys now need a commit. That is the direction #206 said this decision would go, and it is the one that makes the tamper-resistance claim true on both platforms rather than on one.


Starefossen:
Adversarial review pass. Four findings, all real, all fixed in `5d8e8da`.

**Clearing the whole `[propose]` section was a silent loosening.** `gh_guard = true` and `git_push_prevention = true` turn a guard ON with no approval, and survive `--no-gh-guard` (`src/config/loading.rs:1538-1556`). Stripping them from an uncommitted file switched a guard off — invisibly, since `proposed_keys` skips tighten-only rows, so `propose_dropped` stayed false and nothing was printed. That is the exact failure mode this PR exists to prevent, introduced by the fix.

`tightenings_only()` now keeps those two and drops everything else, destructuring `ProposeSection` exhaustively so a new field must be classified rather than defaulting to "kept". The test iterates `PROPOSE_BOOLS` and asserts each row survives if and only if `tighten_only` — read from the registry, not a second hand-maintained list. `gh_guard = false` is a grant and is dropped.

**A named root's uncommitted `.cplt.toml` became a launch-stopper.** An unparseable one is fatal by design, and reading the working tree on Linux is new in this PR — so a previous session could write `&lt;named-root&gt;/.cplt.toml` (agent-writable, not deniable under Landlock), hide it with `.git/info/exclude`, and stop every later launch with a file the user cannot see in `git status`. Fatal now applies to committed files only; an uncommitted one warns and says who probably wrote it. Nothing an author wrote is dropped — there is no committed version to drop.

**`deny.env` could name cplt's own variables.** The sweep runs after cplt sets up the child's environment, so `PATH` (carrying the guard shims), `__CPLT_TRUST_LOCKED` and the proxy variables were deniable: removing a protection rather than a secret. Now refused with the reason. Pre-existing, but this PR is what puts an agent-written file on that path on Linux.

**`config show` still printed "No additional permissions requested."** for a stripped file — the same false line fixed in `trust_show`.

One thing the review did not flag that its first finding created: the notices said the `[propose]` section was ignored "entirely", which stopped being true once tightenings survive. They now say the file grants nothing and name what still applies.

Also cleared: `RepoConfigState::Uncommitted`'s "Never read by cplt on Linux", the dangling "per-root approval waits on #206" in SECURITY.md, and an unqualified "writes are kernel-denied" in the paragraph that now says the working tree is read.

Starefossen:
All three fixed.

**"Commit it" outside a git repository** — right, and that case reaches the same code, since the working-tree load needs no git at all. The notice now branches on `repo_config_state`: a non-git directory is told cplt grants only what a commit contains, without being told to commit into a repository that does not exist.

**Unqualified write-deny in SECURITY.md** — correct, and it matters more than usual here, because that gap is *why* an uncommitted file grants nothing on any platform. The sentence now says: kernel-denied on macOS, and under bubblewrap for a file that exists at launch; not denied at all on the Landlock-only path, since Landlock cannot subtract one file from a writable tree.

**Ignored git exit codes in the fixture** — fixed for both. A fixture that half-applies makes the test assert something other than what it names: had `git rm --cached` failed, the repo would still have had a committed `.cplt.toml` and the test would have been exercising the committed path while claiming to test the uncommitted one.

Issue #281, opened 2026-09-04
CodeQL: rust/log-injection fires on assert! messages in test modules


Starefossen:
Every `assert!` in `src/proxy.rs`'s test module that formats a test-constructed hostname into its failure message trips `rust/log-injection` at severity `error`. The `code_quality: errors` ruleset turns that into a hard merge block, so PR #270 stalled behind 22 of them and had to be cleared by hand. Alerts 40 and 41 were the same thing a few days earlier.

The pattern is a false positive by construction. The "user-provided value" is a string the test wrote two lines up, and the "log entry" is a panic message that only ever reaches a developer reading test output. Nothing crosses a trust boundary.

What makes it worth fixing rather than dismissing again: the trigger is *writing a test assert with a formatted value in it*, which is something we do constantly. Every future PR touching proxy tests inherits the same block, and each one costs a round of manual dismissals before it can merge.

Options, roughly in order of how much they give up:

1. A `.github/codeql/codeql-config.yml` with a `query-filters` entry excluding `rust/log-injection`. Simple, but it drops the rule for production code too, and the proxy is exactly where a real log-injection bug would matter.
2. Same config, but paired with a check that the rule still has somewhere to fire. CodeQL has no path-scoped query filter for Rust inline test modules, so this needs thought.
3. Leave the rule on and keep dismissing. Cheap per incident, unbounded over time.

I lean towards 1 only if we can convince ourselves the proxy's real logging paths are covered by something else, and towards 2 otherwise. Worth a look at whether the alerts can be scoped by the `#[cfg(test)]` boundary at all.

Starefossen:
Recurred within a day, which is the case for fixing this properly rather than dismissing each batch.

Four new alerts (80-83) fired on `src/proxy.rs` at lines 3080, 3279, 3335 and 4378 — all `assert!` failure messages inside the test module, all severity `error`, all blocking every open PR through the `code_quality` rule until dismissed. Dismissed with the same reason as 58-79.

What triggered them: `proxy.rs` was refactored twice today (#304 and #316), and the new tests those PRs added interpolate a hostname into their assertion messages. That is the ordinary way to write a useful test failure in this codebase, so the alerts will keep arriving at roughly the rate we add proxy tests.

Running total is 26 dismissals of this rule, in three batches, across two days.

[...]

PR #495, opened 2026-09-10
feat(link): name the repository, let cplt find and verify the checkout

Starefossen:
The first piece of #491, and the one its review confirmed can land before #206 — user-initiated, reads no `.cplt.toml`, writes no trust entry. (#206 has since landed anyway, in #494.)

```
$ cplt link navikt/sykepenger-model
[cplt] navikt/sykepenger-model → /Users/you/src/sykepenger-model (beside this repository, origin verified)
```

Today the user has to work out the path themselves and remember `cplt config set --local sandbox.repo_dirs`. What this adds beyond convenience is the check that the directory *is* the repository.

## Where it looks

1. Beside the launch repository — one `stat`, and the overwhelmingly common layout.
2. Across the forge directory, for `<host>/<org>/<repo>` (the Go layout): this is what makes `nais/unleash` findable from `navikt/cplt`.
3. The origin of each sibling directory, capped at 50 — a clone in a renamed directory.

`$HOME` at large is never scanned, and a parent that is itself an unsafe root is skipped rather than walked. A directory **inside** the launch repository is never a candidate: that tree is agent-writable, so a checkout planted there is a repository of the agent's choosing wearing the right name.

## Why verification, and what it is worth

A matching name is not identity. Someone with an unrelated `~/src/foo` must not have it linked when a project names `navikt/foo` — that would be a project-grade **write and execute** grant over the wrong tree. So every candidate has `remote.origin.url` read by the trusted git **in the unsandboxed parent** (`detect_current_repo`, which strips `GIT_*` and ignores global config), compared with `repos_match` so `owner/name` is case-insensitive the way GitHub is.

The bound is stated in the docs rather than left implied: on Linux an agent can rewrite `.git/config` in any tree it was granted and Landlock cannot carve that file out of a writable root, so **"origin verified" describes the moment the command ran**. It is not a continuing property of the tree — which is why the link is stored and re-validated as a path, never re-resolved from the name.

## Two things the design review forced, and one reality did

- **Collect every candidate before deciding.** Stop-at-first lets a tree that verifies at an early step silently beat the real repository found at a later one.
- **Validate the identity as `<owner>/<name>` before building paths from it.** User input today; in #491 it is text a committed config chose, and `repos = ["../../.."]` would otherwise have cplt run git in a directory that file picked.
- **Rank the evidence.** Running this against my own machine found 34 linked worktrees of `navikt/copilot`, all reporting the same origin, because worktrees share one config. Pure "refuse any ambiguity" makes the command unusable for anyone who uses worktrees. So: a directory *named* after the repository outranks one matched only by origin; equally good candidates are still refused and both named; worktrees of the winner are counted, and a genuinely separate clone is printed so the operator can link that one instead.

## Verification

Nine unit tests, each a scenario rather than a restatement: sibling found; matching name with a **different** origin not found; cross-org via the forge directory; renamed clone by origin; equally good candidates refused; weaker match reported but not blocking; worktrees counted not listed; a candidate inside the launch repository never taken; a user-named directory still verified, with the error naming what the directory actually is.

Manually against real checkouts: `navikt/copilot` (sibling, 34 worktrees counted), `nais/unleash` (across the forge dir), a non-existent repo (bounded "looked at" list), `../../etc` (refused as an identity), and `--unlink` round-tripping both.

`cargo test`, clippy, and `--target x86_64-unknown-linux-gnu --all-targets` clean.

## Not in this PR

`[propose] repos` — a repository proposing links by identity — which is the rest of #491 and carries the trust questions its review recorded. This is the resolver those need, landed on its own so the search order can be wrong in public before a repository's config depends on it.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -231,9 +231,13 @@ Where it looks, in order: beside the launch repository; then across the forge
 directory, so `nais/unleash` is findable from `navikt/cplt` in a
 `<host>/<org>/<repo>` layout; then the origin of each directory beside the
 launch repository, capped, which finds a clone in a differently-named
-directory. A directory inside the launch repository is never a candidate — that
-tree is agent-writable, so a checkout planted there would be a repository of
-the agent's choosing wearing the right name.
+directory. A directory that resolves inside the launch repository is never a
+candidate — that tree is agent-writable, so a checkout planted there would be a
+repository of the agent's choosing wearing the right name. That check runs on
+the resolved path, so a symlinked sibling pointing back into the project does
+not get around it. Note the sibling directory itself is only as trustworthy as
+your machine: one `--project-dir <parent>` launch makes every directory beside
+the repository writable from the sandbox.
 
 Every candidate has its `origin` read by the trusted git **outside** the
 sandbox and must match the name you asked for. A directory called
@@ -251,7 +255,13 @@ instead. Linked worktrees of the chosen checkout are counted, not listed — the
 share one `.git`, so they all report the same origin.
 
 `cplt link <owner>/<name> --unlink` removes it again, so a root can be dropped
-by the name it was added under.
+by the name it was added under. It matches on the recorded directory's origin
+where it can read one, and falls back to the directory's own name where it
+cannot — a deleted directory has no identity to read, and neither does one whose
+`.git/config` was rewritten since. Without that fallback the entries most worth
+removing would be the ones the command refused to find. If more than one entry
+answers, it names them and removes none; `cplt config set --local
+sandbox.repo_dirs <DIR> --unset` is always available.
 
 What this writes is an ordinary `sandbox.repo_dirs` entry, so everything below
 applies to it unchanged. What it adds is the identity check. And note the
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -82,6 +82,9 @@ cplt --repo-dir libs/sykepenger-model exec -- ./gradlew build
 
 # Persisted for this checkout, so every later launch picks it up with no flag
 cplt config set --local sandbox.repo_dirs ~/src/spleis/libs/sykepenger-model
+
+# Or name the repository and let cplt find and verify the checkout
+cplt link navikt/sykepenger-model
 ```
 
 The startup summary then lists what is in scope, and where each entry came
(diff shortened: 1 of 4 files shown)


Starefossen:
Review pass. Seven findings, all real, fixed in `f3508e3`.

**The containment check was defeated by a symlink** — the serious one, and reproduced before fixing:

```
$ cplt link navikt/sykepenger-model
[cplt] navikt/sykepenger-model → .../spleis/planted (beside this repository, origin verified)
```

`dir.starts_with(project_dir)` ran on the path as written, so `../sykepenger-model -> <project>/planted` passed it, canonicalized back into the project tree, verified against a planted origin and took rank 0. cplt then vouched for a tree the agent authored while the user believed the real repository was in scope. The check moved inside `verify`, after `canonicalize`; `verify_named` got it too, since the explicit form must not accept what the search refuses.

**And the test for it passed with the guard deleted.** Correct: no search step generates a path inside the project, so a plainly-nested decoy is never a candidate and the test could not fail. It is the symlink case now, and falsified — removing the check fails it. Worth stating plainly, because a test that cannot fail is worse than no test: it is a claim of coverage.

**A planted `.git` file reads as a worktree of the real repository.** Whether the directory's `.git` is its own is now checked before the name-vs-origin ranking, so the main checkout outranks both worktrees and that decoy.

**`--unlink` could not remove what most needs removing.** Origin-only matching means a deleted directory (nothing to read) and a tree whose `.git/config` an agent rewrote both get "nothing linked here is navikt/foo" — while the write+exec grant stays. It falls back to the recorded directory's name, which only ever *removes* a grant and prints what it removed, and the refusal now lists what is actually linked. `choose_unlink` is a pure function taking the identity reader, so the four cases are tested without a git tree — it had no test at all before.

**The origin string was printed unsanitized.** `git config --get` returns repo-controlled bytes verbatim, escapes included, and it went into a suggested command. Quoted unless it parses as an identity.

Plus the two small ones: the no-trusted-git message suggested naming the directory, which needs the same git; and the capped scan reported how many directories it read without saying it stopped early.

Docs updated on both counts the review called overclaims: `--unlink` carries the same "moment you ran it" bound, and the sibling directory is only as trustworthy as the machine — one `--project-dir <parent>` launch makes every sibling writable from the sandbox.

Starefossen:
Three of the four threads addressed; one was already fixed.

**Symlink bypass of the containment guard** — fixed before this comment thread was posted, in `f78d4e2`: the check moved inside `verify()`, after `canonicalize`, and `verify_named` got it too. Reproduced first, and the test that covers it is the symlink case, since a plainly-nested decoy is never generated by any search step.

**`--unlink` did not validate the identity** — right, and it printed the string back to the terminal. Same validation as the other two forms now.

**`load_local` errors swallowed by `.ok().flatten()`** — right, and the consequence is the wrong direction: an unreadable local config reported as "nothing is linked", when the launch fails loudly on the same file. It now fails loudly too.

**CodeQL `rust/cleartext-logging`** — dismissed as a false positive (alert 88). The value traced is the path of the trusted git binary, which `link::resolve` uses to run `git rev-parse` and `git config --get remote.origin.url`. What actually reaches the terminal is a count of linked worktrees and the directory the user just asked cplt to find. Nothing on that path is a credential, and the git binary path is not sensitive — `cplt doctor` prints it deliberately. Same shape as the log-injection dismissals tracked in #281: a taint label surviving a function whose output is unrelated to the tainted input.

PR #496, opened 2026-09-10
feat(repo-config): a project may propose the repositories it spans, by name

### LINKED ISSUE #491: Design: propose linked repositories by identity, resolve and verify them locally

Starefossen:
Closes #491, on top of the resolver that landed in #495.

```toml
[propose]
repos = ["navikt/sykepenger-model", "navikt/spleis-testdata"]
```

```
$ cplt trust accept repos
[cplt] navikt/spleis-testdata: not linked
  No checkout of that repository was found. Looked at: ...
```

## The shape of it

An **identity, never a path**. The repository states what the project is; the user's machine decides where those repositories live. That is what makes this allowable at all — the refusal it appears to contradict ("repo config cannot grant paths") is about paths, and `navikt/foo` is not one.

Approving is where names become paths, and it is the only place. Each identity is resolved and origin-verified by the same code `cplt link` uses. A repository that is not on the machine is reported, never fetched.

## Three consequences of approval being an *action*, not a value

- **`cplt trust revoke repos` unlinks what it linked.** The launch reads `sandbox.repo_dirs` and never consults the trust store, so without recorded provenance this would be the only proposal whose grant survives its own revocation. The approval records `{identity, path}` pairs and revoke removes exactly those — a root the user added with `cplt link` or `config set --local` is theirs and stays. Its adversarial review raised this on the design; it is the load-bearing piece of the implementation.
- **`--accept-repo-config` cannot approve it.** One-run flag, persistent grant: that pairing is #206's bypass class, and the reason this feature waited for #206 rather than merely citing it.
- **Re-approving re-links.** "Already approved" is not "already done" when approval is an action. A user who approved before cloning the repository, or who removed a root by hand, must be able to run it again. I found this because a test failed exactly that way, on a fixture that had an approval from a previous run.

## Also

The identity is validated in `validate_repo_config`, where the file is read, not at the resolver: the value reaches path construction on the user's machine and it is text a committed config chose. `repos` is a grant, so `tightenings_only` drops it — an uncommitted `.cplt.toml` cannot propose repositories, per #494. And the unapproved-permissions warning names the repositories, because `repos` as a bare key says nothing about what approving it would put in scope.

## Verification

Unit: identities not paths (six malformed forms refused), and `repos` dropped from an uncommitted file.

e2e: the full loop — unapproved grants nothing, approval links the proposed repository and leaves a hand-added root alone, revoke removes the approval's root and keeps the user's. Falsified by deleting the unlink call.

e2e: `--accept-repo-config` does not link, **and** the run still reports `repos` as unapproved. The first assertion alone was vacuous — approving `repos` for one run does nothing observable, since linking only happens in `trust accept` — so the test passed with the filter removed. The second assertion is what actually fails without it.

Manually, on scratch repositories: propose two repositories where one exists, approve, check the startup summary lists it as a new read/write/exec tree, add a hand-linked root, revoke, confirm only the approval's root went.

`cargo test`, clippy and `--target x86_64-unknown-linux-gnu --all-targets` clean.

## One thing deliberately not built

The design floated resolving proposals at launch to show "found at ../x, not yet linked". That costs a directory scan on every start to print a line, so the summary names the repositories without resolving them. Worth revisiting only if someone finds the unresolved line too vague.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -216,6 +216,56 @@ table.
 The long forms are used throughout this document because they say which file is
 being written.
 
+### A project can name the repositories it spans
+
+A `.cplt.toml` may say which repositories a session usually needs:
+
+```toml
+[propose]
+repos = ["navikt/sykepenger-model", "navikt/spleis-testdata"]
+```
+
+Identities, never paths. The repository states what the *project* is; your
+machine decides where those repositories live. A committed `.cplt.toml` cannot
+point cplt at a directory of its choosing — the most it can do is put a name in
+front of you.
+
+Approving is what turns names into paths, and it is a separate, explicit step:
+
+```
+$ cplt trust accept repos
+[cplt] sandbox.repo_dirs = [/Users/you/src/sykepenger-model]
+[cplt] navikt/spleis-testdata: not linked
+  No checkout of that repository was found. Looked at: ...
+```
+
+Each identity is resolved and origin-verified exactly as `cplt link` does, and
+one that is not on your machine is reported, never fetched: cloning a
+repository named by a config file is a much larger decision than linking one
+you already have. Approving again re-links anything missing, so a repository
+you clone later needs no re-approval dance.
+
+`cplt trust revoke repos` removes the roots that approval created — and only
+those. A root you added yourself with `cplt link` or `config set --local` stays,
+because it is yours.
+
+Three properties worth knowing:
+
+- **An unapproved proposal grants nothing.** It shows up in the launch summary
+  naming the repositories, so you learn the link exists to be made, and that is
+  all it does.
+- **`--accept-repo-config` cannot approve it.** That flag approves for one run
+  and persists nothing; linking a repository is persistent. It is the one
+  proposal the flag skips.
+- **An uncommitted `.cplt.toml` cannot propose repositories at all.** It is a
+  grant, so it needs the audit trail a commit gives — see the repo-config
+  security model above.
+
+The approval binds to the path it resolved, which is then re-validated on every
+launch like any other named root. cplt does not re-resolve the identity each
+time: a later directory rename would otherwise redirect a grant you approved
+for a specific tree.
+
 ### Naming a repository instead of a path
 
 `cplt link` takes the repository, finds the checkout, and checks it really is
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -230,7 +230,10 @@ machine decides where those repositories live. A committed `.cplt.toml` cannot
 point cplt at a directory of its choosing — the most it can do is put a name in
 front of you.
 
-Approving is what turns names into paths, and it is a separate, explicit step:
+**What approving grants is not small.** Each linked repository becomes a named
+root: read, write **and execute** on that whole tree, the same posture as the
+project directory. A repository proposing five repositories is asking for five
+of those. Approving is therefore a separate, explicit step:
 
 ```
 $ cplt trust accept repos
@@ -245,9 +248,14 @@ repository named by a config file is a much larger decision than linking one
 you already have. Approving again re-links anything missing, so a repository
 you clone later needs no re-approval dance.
 
-`cplt trust revoke repos` removes the roots that approval created — and only
-those. A root you added yourself with `cplt link` or `config set --local` stays,
-because it is yours.
+`cplt trust revoke repos` removes the roots recorded at approval time, matched
+by path — and only those. A root you added yourself with `cplt link` or
+`config set --local` stays, because it is yours: if a proposal names a
+repository you had already linked, approving says so and leaves it alone rather
+than taking ownership of it. `cplt trust` lists what an approval linked, which
+is what revoke acts on; a recorded root that is no longer in your config is
+reported rather than assumed gone, because revoke cannot tell removed from
+renamed.
 
 Three properties worth knowing:
 
@@ -265,6 +273,21 @@ The approval binds to the path it resolved, which is then re-validated on every
 launch like any other named root. cplt does not re-resolve the identity each
 time: a later directory rename would otherwise redirect a grant you approved
 for a specific tree.
+
+**"Origin verified" describes the moment you approved.** On Linux an agent can
+rewrite `.git/config` in any tree it was already granted, and Landlock cannot
+carve that file out of a writable root — so a tree a previous session could
+write can claim to be the repository a proposal names. If the real checkout is
+absent, that claim is what gets linked, and the verification line will say so.
+It grants nothing the previous session did not already have, but it can make
+you believe the real repository is in scope when it is not. This matters more
+here than for `cplt link`, because what is being verified was chosen by a
+committed file rather than typed by you.
+
+Limits worth knowing: a second checkout of the same repository cannot approve
+`repos` at all, because trust is stored per origin; a fork reads as "not found",
+since its `origin` is the fork; a repository with no `origin`, or with only
+`upstream`, is invisible; and a monorepo can only be named whole.
 
 ### Naming a repository instead of a path
 
--- a/src/link.rs
+++ b/src/link.rs
@@ -600,6 +600,15 @@ mod tests {
             "..",
             "nav ikt/cplt",
             "navikt/./cplt",
+            // Two components, and one of them walks up. `name_candidates` joins
+            // the name onto the parent directory, so this is the shape that
+            // reaches `parent.join("..")` — and it is the one the earlier list
+            // missed, because every other bad case was refused by the arity
+            // check alone (#496 review).
+            "navikt/..",
+            "../cplt",
+            "./cplt",
+            "navikt/.",
         ] {
             assert!(!is_valid_identity(bad), "{bad:?} must be refused");
         }
--- a/src/main.rs
+++ b/src/main.rs
@@ -2440,7 +2440,7 @@ fn resolve_context(cli: &Cli, check_mode: bool) -> anyhow::Result<ResolvedContex
             unapproved_proposals.len()
         ));
         for key in &unapproved_proposals {
-            match propose_key_detail(&repo_propose, key) {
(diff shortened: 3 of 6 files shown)


Starefossen:
Adversarial review pass. Eight findings, all real, fixed in `97f525f`. Three of them were the same failure: a read/write/execute grant left standing with no command able to withdraw it.

**The record of what an approval linked was replaced, not merged.** That record is the only route `trust revoke repos` has to those roots — the launch reads `sandbox.repo_dirs` and never consults the trust store. Three ordinary sequences dropped a record while its root stood:

- the project drops an identity from `[propose] repos` and the user re-approves;
- an identity stops resolving (directory moved, or a second clone makes it ambiguous);
- the hash goes stale and the user approves *some other key* — `carried` is `None` there, so the fresh entry began with an empty list.

Reproduced the third end to end. Records now merge, and survive a stale hash: "re-review the values" is not "those roots stopped existing". Bound to the checkout the same way approvals are, so a foreign entry's roots are never adopted.

**An already-linked root was claimed by the approval.** A user runs `cplt link navikt/model`; the project happens to propose the same repository; `trust revoke repos` then deletes the user's root. Directly against what the docs promise. It is now left as theirs, with a test that falsifies.

**Revoke compared paths as written** while the local layer stores `~/…` verbatim and an approval records the canonical path — so a hand-edited config left the grant in place while revoke printed success. Compared expanded now, and a recorded root missing from the config is *reported*, not assumed gone: revoke cannot tell removed from renamed, and asserting the safe reading is how a live grant gets announced as withdrawn.

**The consent prompt truncated the list it was asking about.** My five-entry cap belongs to the launch warning, not to the question ending in `[y/N]` — 200 trees approved while showing 5. Worse, the "see `cplt trust`" it pointed at never listed them: `trust show` prints key names only. It now prints the roots an approval created, labelled read+write+exec, which is also what revoke acts on.

**`revoke repos` unlinked before the `removed == 0` early return**, which skips the save: roots gone from the config, still recorded as linked, put back by the next accept.

**The re-link branch undid deliberate unlinks.** "Approved before the repository was cloned" and "the user removed this root on purpose" are different cases, and the record distinguishes them — only identities with no record are linked there.

Two smaller ones: `repos` was the only list not sorted in the hash normalizer, so a pure reorder invalidated an approval (and could trigger the first finding); and the identity test's six bad cases were all refused by the arity check alone, so the `..` rules were untested — `navikt/..` and `../cplt` are covered now, falsified by removing the rules.

Docs: what approving grants was never stated (read, write **and execute** on a whole other repository — it was in a source comment only); the Linux "origin verified describes the moment you approved" bound now appears in this section rather than ten lines below, since here the thing being verified was chosen by a committed file rather than typed by the user; and the fork, second-checkout, no-origin and monorepo limits are stated.

Not changed: the review's point about no bound on how many repositories one approval links. It is user-initiated and now fully displayed before the prompt, so the consent problem is gone and what remains is cost the operator can see.

Starefossen:
Already fixed — this one was posted against the earlier head. `link_proposed_repos` now takes the `already` branch and `continue`s without recording:

```
• navikt/model → /Users/you/src/model (already linked; left as yours)
```

`e2e_approving_does_not_claim_a_root_the_user_linked_first` covers it: the user links, the project proposes the same repository, approve, revoke, and the root is still there. Falsified by restoring the fall-through — it fails.

PR #497, opened 2026-09-10
fix(repo-config): an unreadable .cplt.toml in a linked repository warns

### LINKED ISSUE #472: A .cplt.toml that cannot be parsed stops the launch in a named repository and only warns in the launch one
Introduced deliberately in #468 and worth settling rather than leaving as an inconsistency for the next reader to trip over.

## The current behaviour

- **Launch repository.** `load_repo_config` returns `Err`, and the launch prints `Failed to load .cplt.toml: <reason>` and carries on. The session runs without whatever `[deny]` that file contained.
- **A `--repo-dir` root.** The same failure is fatal: the launch stops, naming the repository and saying its `[deny]` section cannot be applied.

## The argument for making both fatal

It is the same argument, and it is the one #468 acted on for named roots. A `.cplt.toml` that *exists* and cannot be parsed means a restriction its owner wrote is missing. The file is committed, so the operator has every reason to believe it is in force. A warning is exactly the wrong shape for a missing restriction: it scrolls past on the way to a session that runs anyway, and nothing later in the session says the deny is absent.

This is also the shape of several fixes already made here — #214's "absence of information rendered as a clean bill of health", and the `unaudited_roots` warning that exists because measuring nothing must not read as measuring clean.

## The argument for leaving the launch repository as it is

It is long-standing behaviour, and changing it is breaking: anyone with a malformed committed `.cplt.toml` currently gets a warning and a working session, and would get a hard stop instead. The parse error may also be in a section that grants nothing — a mangled `[propose]` block with an intact `[deny]` is indistinguishable from the other way round, because the parse fails whole-file.

There is a real asymmetry in the opt-in, too: a named root is something the user chose to add to this session, and holding an explicitly-added repository to a stricter standard is defensible on its own.

## What would settle it

Whether any repository in the org currently has a `.cplt.toml` that fails to parse. If none does, the breaking change costs nothing and the consistent rule wins. If some do, the fix is to make it fatal *behind the staged rollout* (#264) rather than immediately.

Related: #206, which is about a different half of the same file's trust model.

Starefossen:
Closes #472, the last inconsistency in this area after #206 (#494).

| | launch repository | linked repository |
| --- | --- | --- |
| before | warns, launch continues | **stops the launch** |
| after | warns, launch continues | warns, launch continues |

## Why warn, rather than make both fatal

The argument for fatal is real and #472 states it well: the file exists, so a restriction its owner wrote is missing, and a missing restriction must not be a warning the operator scrolls past. But that argument applies just as much to the launch repository, which has always warned — so the defect was having two answers to one question, not either answer alone.

Making both fatal would stop a session over a typo in a config belonging to a repository the user is not working in. A linked repository is a convenience for reaching a library beside your project; a broken file there should not be able to prevent you from starting work in your own repository.

What makes warning defensible rather than a hole: **a named repository's file grants nothing.** Its `[propose]` is never consulted, so the only thing that can go missing is a tightening. And the warning says what that costs, in terms of what the session can now do:

```
[cplt] Cannot read .cplt.toml in the linked repository ~/src/model: Invalid .cplt.toml: ...
  Its [deny] section is NOT applied, so this session is less restricted inside that
  repository than its config asks for. Fix the file, or drop the repository from the
  linked set.
```

## It also closes a denial-of-service that #494 only half-closed

A named root is agent-writable, Landlock cannot deny `.cplt.toml` inside it, and `.git/info/exclude` hides a new file from `git status`. Fatal handed a previous session a way to stop every later launch with a file the user cannot see. #494 fixed that for an *uncommitted* file; a committed one had the same shape wherever an agent could commit.

## Verification

`an_unparseable_cplt_toml_in_a_named_repository_stops_the_launch` — the test that pinned the old behaviour — failed on this change, which is exactly what it was for. Rewritten as `..._in_a_linked_repository_warns_and_the_launch_continues`, asserting the launch succeeds **and** that the cost is stated (`[deny] section is NOT applied`, `less restricted`), rather than matching the exact phrasing. Falsified against a restored `bail!`.

`cargo test`, clippy and `--target x86_64-unknown-linux-gnu --all-targets` clean. SECURITY.md and `docs/configuration.md` updated — both previously documented the fatal rule and its reasoning.


Starefossen:
Right, and it was my doing: `cplt link` (#495) introduced a second vocabulary for something the codebase already had a word for.

Settled by deciding which is the noun and which is the verb. **A named repository is the thing** — what `--repo-dir` and `sandbox.repo_dirs` produce, with the read/write/execute grant. **Linking is one way to name one**: the way that does not require knowing where it lives. So messages use the noun, and both docs sections now say so explicitly rather than leaving a reader to work out that `cplt link` and `--repo-dir` produce the same thing.

Warning text, test name and doc wording all updated.

Issue #386, opened 2026-09-06
PRD: macOS/Linux behaviour parity — close every divergence that is not a kernel limit


Starefossen:
cplt behaves differently on macOS and Linux in ways that are not all forced by the kernels. Some divergences are genuine mechanism limits, some are code gaps, and some are choices we made and then argued about internally. This PRD separates those three and proposes how to close everything in the second and third categories.

cplt runs on macOS and Linux only. There is no Windows port (no `target_os = "windows"` anywhere; CI builds two Apple and two Linux targets). WSL2 is supported as Linux, with a guard that treats a Windows-side `gh`/`git`/`node`/`npm` resolved over interop as a hard failure.

## The principle to adopt

**The same configuration must produce the same outcome on both platforms, or cplt must refuse to run.**

`AGENTS.md` already states this for grants: *"A grant or setting that cannot be honoured must fail loudly at the point it is written. It must never be accepted and then quietly dropped."* A **deny** that cannot be honoured is the same class, and today it does not follow the rule — `--deny-path` on a Landlock-only host prints `File deny paths were skipped` and continues with strictly less protection than the user configured.

Adopting the principle means "different" always means "the kernel cannot" and never "we chose to".

## The one fact that explains most of the divergence

Seatbelt can deny a subpath inside an allowed tree. **Landlock cannot** — its rules are purely additive with no deny, so once a tree is writable nothing inside it can be clawed back. That single asymmetry produces most of the ❌ column in the platform comparison table.

Bubblewrap restores it with read-only bind mounts. So the honest framing is:

> **Linux reaches macOS parity only with bubblewrap. Without user namespaces, several controls cannot exist.**

That gives a clean decision rather than a long tail of special cases: either require bubblewrap when a control needs it, or refuse the control and say so.

## Workstream 1 — make the response to an unenforceable control uniform (highest value)

Today an unenforceable deny warns and continues. Proposed: **fail closed by default, with an explicit opt-out.**

- When `deny.paths` / `--deny-path` is configured and the backend cannot enforce it, refuse to launch. The message should name the reason (Landlock-only host, no user namespaces) and the two ways forward: install/enable bubblewrap, or acknowledge the degradation explicitly.
- Add one opt-out (`--allow-unenforceable-denies`, or a config key) so a user who genuinely wants best-effort can have it, once, visibly. Same shape as the existing dangerous-grant opt-ins.
[...]

Issue #344, opened 2026-09-04
PRD: Sessions that span several repositories


Starefossen:
All line numbers are against `c941a3b` (2026-09-04).

## Recommendation in one paragraph

Keep one launch repository as the session's identity and let the user name additional repositories explicitly. Do it in two stages that each stand alone. **Stage 1** adds no flag: it makes the sibling you can already grant with `allow.write` visible and truthful (startup summary, brief, guard messages, `cplt check`, docs) and closes the two holes that already bite at N (#313, the `allow_push` bare-name fallback). **Stage 2** adds `--repo <dir>` (repeatable): a second project-grade root that can build and test, whose `[deny]` is honoured, that `gh` and `git push` can target under the same rules as the launch repo, and that the audit covers. No policy merging, no workspace identity, no `[propose]` from a second repo until #206 is decided.

The honest finding: for "edit the library while building the consumer", `--allow-write ../lib` already works and the gap is that nothing tells the user or the agent what that grant is and is not. For "edit *and build* both" there is no path today short of two terminals, because a write grant is deliberately non-executable on both backends and `allow.exec` refuses to overlap it. That is the one capability Stage 2 adds.

## #165 and #216 status

**#165 is superseded by this PRD and should be closed.** Its merge model cannot be built: relative allow-lists anchor to their own repo and drop escapes (`src/config/loading.rs:1223-1240`), so intersecting two repos' allow-lists is empty by construction; repo config is additive-only (`src/config/loading.rs:1086-1196`, `src/config/repo.rs:44-119`, booleans only go `false → true`); `[propose]` has mixed polarity (`gh_guard`/`git_push_prevention` enable a guard). None of its acceptance criteria exist: no `--workspace` flag, no `[workspace]` key anywhere in `src/`. Its short-term docs bullet is folded into Stage 1 below, with one of its recommendations reversed (never start from a parent directory).

**#216 is the right analysis and is retained as background; four of its statements need correcting, and this PRD replaces its staged plan:**

1. *Stage 0 is done.* #212/#214 shipped in #217, #213 in #230, #215 in #289.
[...]

PR #498, opened 2026-09-10
fix(bubblewrap): protect repositories nested inside a writable root

Starefossen:
Closes the last unticked box in #344, and the first code-gap bullet in #386's workstream 2.

## The gap

macOS expresses the `nested` entries of `PROTECTED_IN_ROOT` as a regex matching any depth. bubblewrap has no regex — it binds paths — so the paths must be enumerated, and they never were. With `--allow-write ~/src`:

| | macOS | Linux (bwrap), before | Linux (bwrap), after |
| --- | --- | --- | --- |
| `~/src/.git/hooks` | denied | read-only bind | read-only bind |
| `~/src/repo/.git/hooks` | denied | **writable** | read-only bind |
| `~/src/repo/.cplt.toml` | denied | **writable** | read-only bind |

A hook planted in that nested `.git/hooks` runs **unsandboxed** on the user's next git operation in that repository. `.cplt.toml` steers the next session's sandbox. Both are the exact vectors the tables exist to close, and both were open one directory deeper than the code looked.

## The bound, and why it is stated rather than hidden

A writable root is arbitrary — it can be a home directory or a network mount — so the scan is bounded: **three levels** below the root, **4096 directory entries** per root, and dependency/build directories skipped by name (a `.git` inside `node_modules` is vendored, not worked in).

Stopping early *under*-protects, which is the wrong direction for a bound to be silent about. Hitting the cap warns, names the root, and points at `--repo-dir` or a narrower grant. The depth limit is documented in SECURITY.md and `known-impacts.md` rather than left for someone to discover — macOS still matches at any depth, and that difference is now written down instead of being a comment in a test.

The nested subset is read from the same `nested` flag on the policy table that macOS reads, so a future nested entry reaches both platforms without a second list to forget.

## Verification, and a caveat about how

`sandbox_bubblewrap` is `cfg(target_os = "linux")`, so this test **cannot run on the macOS host I am on**. I ran it in a Linux container instead:

- 41 bubblewrap tests pass.
- The new test — `a_repository_nested_in_a_writable_root_gets_the_same_protection` — fails when the nested loop is stubbed to empty.
- It also asserts the bounds hold: a repository four levels down and one inside `node_modules` are **not** bound, so the limits are pinned rather than assumed.

Two landlock tests (`generate_policy_*`) fail in that container on unmodified `main` as well — container-specific, not this change.

macOS suite, clippy, and `cargo check --target x86_64-unknown-linux-gnu --all-targets` all clean.


Starefossen:
Review pass, all measured in a Linux container. Three findings were ways around the protection this PR adds, and one of my tests was vacuous.

**The binds had no pins** (finding 1). A read-only bind pins content, not the name — `mv .git g2 && mkdir .git` hands back a writable `.git/hooks` under a name no bind covers. `rename_pin_paths` closes that for a root, and the nested set never reached it, so the feature promised more than it delivered: GHSA-39xf-9j26-f82m one directory down. The scan now runs **once** and feeds both sets, which also removes the double walk.

**And my test for it could not fail.** It called `rename_pin_paths` directly, so it passed with the wiring deleted — the helper was never the broken part. The assertion moved onto the assembled sets in `sandbox.rs`, verified by deleting each call in turn.

**The cap was steerable** (finding 2). Files spent budget and the root's own `read_dir` drained first, so 6000 junk directories found *zero* repositories. And at 4096 it tripped at ~150 repositories — with `~/go/src` putting every repository at exactly the depth truncation drops (finding 4). Now only directories count, the walk is breadth-first with sorted entries so the same tree yields the same set on every launch, and the limit is 20000 directories — about 40 ms, against a failure mode that is lost protection.

**Symlinked directories were followed** (finding 3). That pulled repositories outside the granted tree into the bind set, and — the sharper half — bubblewrap refuses to mount onto a symlink, so one planted inside a granted tree fails the whole wrapper and downgrades the next session to Landlock-only. `entry.file_type()` does not follow.

Smaller ones: a skipped *name* is still tested for being a repository (`~/src/dist` may be a checkout, finding 6); the cap warning fires only when bubblewrap is active, since on a Landlock-only host it implies a protection that does not exist (finding 7); the hardcoded 4096 in the message is gone.

**Bare repositories and nested worktrees/submodules** (finding 5): the code comment claimed `git_dirs` covered them and it does not — that only carries the launch repository and named repositories. The claim is deleted and the gap is recorded in SECURITY.md instead. Resolving gitdirs per nested repository is a real follow-up, not something to assert away here.

Docs now state the bounds *as part of the security claim* rather than as trivia, including the 20000-directory cap the review rightly said matters more than the depth bound.

Five new tests, each falsified: nested binds, the pins wiring, the depth boundary at the exact level, the symlink guard, and the cap being reported. Two `generate_policy_*` tests fail in the container on unmodified `main` too — container artefacts.

Starefossen:
One stale, one half-right — thanks for both.

**`path.is_dir()` following symlinks** — fixed before this landed, in the review pass. `entry.file_type()` does not follow, which closes the escape you describe and the loop/budget-burn with it. There is a sharper half too: bubblewrap refuses to mount onto a symlink, so a symlinked bind *destination* fails the whole wrapper and downgrades the session to Landlock-only. `a_symlinked_directory_is_not_walked` covers it, falsified by putting `path.is_dir()` back.

**The cap warning** — right on both counts, now fixed. The number had its own copy in the message, and a message with its own copy of a constant is one that will be wrong after the next change; `NESTED_SCAN_LIMIT` is the single definition and the warning formats it. The wording named only `.git/hooks`, when a repository past the cap gets *none* of the nested protections — `.cplt.toml`, `.github/hooks` and the agent auto-exec paths all stay writable there. It says that now.

Issue #306, opened 2026-09-04
cplt config set accepts a grant the tool refuses at every launch


Starefossen:
Found while investigating #291, by accident — which is the argument for fixing it.

`cplt config set allow.read ~/.netrc` succeeds. It writes the entry, prints the updated list, and exits 0. The path is in `DENIED_FILES`, so `prepare()` refuses it at the next launch, and at every launch after that.

So the tool accepts a setting at write time that it rejects at read time. The user learns about it only when they next start an agent, from a message about a line they added at some earlier point and have probably forgotten.

## Why this is worth closing

#207 deliberately made a hard-denied grant a startup **error** rather than a silently dropped rule, on the reasoning that silently ignoring a line the user wrote is how people come to believe they have access they do not. That reasoning applies one step earlier too: the moment to say "this cannot be granted" is when someone types it, not on every subsequent run.

The write path already has everything it needs. `policy::hard_denied_file(home, path)` exists and is what `prepare()` consults; `cplt config set` could call the same predicate and refuse with the same message.

## Scope question worth settling

Whether `config set` should also refuse the *other* grants that `prepare()` rejects, rather than just `DENIED_FILES`. As of #295 an exec grant overlapping a writable tree is refused at resolve time, and there may be more of these as validation grows. Sharing one validation entry point between `config set` and `prepare()` would keep the two from drifting, which is the actual defect here — the same question answered in two places, differently.

Related: #207 (hard-denied grants became a startup error), #291 (`~/.ssh` directory grant, silent on macOS).

Starefossen:
Closed by #501: `config set` and the launch share one predicate now (`validate_grant_path`), so a check added for either reaches both — which was the actual defect this issue named, rather than the individual message.

`allow.exec` gets the part of the overlap check a single grant can be judged against; an overlap with an `allow.write` added later still surfaces at launch, where the resolved config exists.

PR #499, opened 2026-09-10
fix(proxy): refuse a list file the sandboxed agent can edit

### LINKED ISSUE #426: The live-reloaded proxy allowlist can be pointed at a file the agent can write
## Summary

`proxy.allowed_domains` names a file that the proxy re-reads every `RELOAD_TTL` (5 seconds) — by design, so that editing it adds and revokes entries live (`src/proxy_domains.rs:19-21, 43`). Nothing checks where that file lives. If it is inside the project directory or any `allow.write` grant, the sandboxed agent can edit its own egress allowlist and see the change take effect within five seconds.

Found while researching whether any sandbox setting could be updated in flight. The answer turned out to be that one already is, and its only channel is one the agent can reach.

## Mechanism

`allowed_domains_file` is taken from resolved config with no location constraint (`src/main.rs:2350-2352`, validated at `:2373` only for readability and parse). The reload is deliberate and documented:

> **Reloadable** entries come from a file that is re-read every `RELOAD_TTL`, so editing that file adds *and revokes* entries live.

That is the right design for a human editing an allowlist mid-session. It is the wrong one when the file is reachable by the process the allowlist constrains.

There is a second path to the same place. `~/.config/cplt` is in `DENIED_DOTFILES` (`src/sandbox_policy.rs:29`), but `emit_denied_dotfile_overrides` (`src/sandbox_profile.rs:1914`) deliberately re-allows an explicitly granted subpath inside a denied dotfile directory — a supported override, unlike `DENIED_FILES` where no grant may re-open the path. So `allow.write = ["~/.config/cplt/<something>"]` also hands over write access to state cplt trusts. The refusal message for granting the parent directory actively recommends this shape:

> Name the specific path you need inside it instead, e.g. `~/.config/cplt/<file>` — that grant works on both.

The same override reaches the trust store and the blocklist subscription cache.

## Impact

Both paths require the user to have configured them — an allowlist file inside the project, or a grant inside cplt's state directory. Neither is a default. So this is a footgun rather than a live vulnerability, and it is filed as an issue rather than an advisory for that reason.

What makes it worth closing anyway: the failure is silent and continuous. A control that polls agent-writable state every five seconds to ask what the rules are is the exact shape of GHSA-cm6f (the guard asking the sandbox which branch was protected) and of GHSA-39xf (a deny whose parent the agent could rename). The pattern has produced four advisories this week; this is the same pattern in the one subsystem that re-reads by design.

## Suggested fix

1. **Refuse an `allowed_domains` path that resolves inside the project directory or any writable grant**, at launch, with a message naming both the path and the reason. This is the direct fix and it is cheap — the writable roots are already computed by then.
2. **Refuse grants inside `config_dir()` as a subtree**, in `validate_hard_denied_grants` (`src/sandbox.rs:272`). No grant inside cplt's own state directory is legitimate, and this also closes the trust-store and blocklist-cache variants. It needs the parent-directory refusal message to stop suggesting a path inside as the remedy.
3. Consider whether the proxy's list files should carry `DENIED_FILES` semantics — absolute, not re-openable by any grant — rather than relying on the configured path being sensible.

Items 1 and 2 are independent and either can ship alone. Item 2 was already identified as stage 1 of the #340 work for the same reason.

Starefossen:
Closes #426.

```
[cplt] proxy.allowed_domains names allow.txt, which is inside ~/src/myproject — a tree this session can write.
  The proxy re-reads that file every few seconds, so the agent could edit its own egress rules and
  have them take effect mid-session. Move the file outside the writable trees (for example
  ~/.config/cplt/), or drop the grant.
```

## Why refuse rather than warn

The live reload is deliberate — editing a list file adds and revokes entries without a restart, which is right for a person and wrong for a file the constrained process can reach. `AGENTS.md` already states the rule for grants: *a grant or setting that cannot be honoured must fail loudly at the point it is written*. An allowlist the agent can edit cannot be honoured, so it fails at launch, the same way `allow.exec` refuses a tree overlapping a writable one.

The issue calls this a footgun rather than a live vulnerability, and that is fair — it needs the user to have configured it. What makes it worth closing is the shape: a control that asks agent-writable state what the rules are, every five seconds, is GHSA-cm6f (the guard asking the sandbox which branch was protected) and GHSA-39xf (a deny whose parent could be renamed) in the one subsystem that re-reads by design.

## Both files

`proxy.blocked_domains` gets the same check. A blocklist the agent can edit is the same control with the sign flipped — it would let a session remove its own blocks.

## Details

Compared on **canonicalized** paths, because the grant and the configured file can spell the same directory differently, and a spelling that slipped past would be the bypass itself. Falsified by removing the canonicalization: the symlinked-grant case fails.

The issue's second path — grants inside cplt's own state directory — is already closed by `cplt_state_dir_grant`, so this PR is item 1 only.

## Verification

`a_proxy_list_file_in_a_writable_tree_is_found` covers the project directory, an `allow.write` grant, a symlinked spelling of that grant, and — the case that must keep working — a file outside every writable tree.

End to end: refused from inside the project, refused via `--allow-write`, and `~/.config/cplt/list.txt` still loads normally (`Domain allowlist: 1 domains from …`).

`project_proxy_allowlist_blocks_unlisted` put its fixture inside the project — the refused shape — so it caught the change. Its list moved to a tempdir, which is what the docs now tell users to do.

### DIFF
--- a/docs/proxy.md
+++ b/docs/proxy.md
@@ -212,6 +212,24 @@ An IP literal cannot be waived — give the host a name. A repository may propos
 entries in `[propose.proxy]`, which is the only proxy key `.cplt.toml` can
 propose, and they apply only after `cplt trust accept`.
 
+### Where the list files may live
+
+A list file is re-read every few seconds, which is what makes an edit take
+effect without restarting the session. That is the right design for a person
+editing an allowlist and the wrong one for a file the sandboxed agent can reach
+— so cplt **refuses to launch** when `proxy.allowed_domains` or
+`proxy.blocked_domains` names a path inside the project directory or any
+`allow.write` grant, naming both the file and the tree:
+
+```
+[cplt] proxy.allowed_domains names allow.txt, which is inside ~/src/myproject
+       — a tree this session can write.
+```
+
+An allowlist the agent can edit is not an allowlist. Keep the file somewhere the
+session cannot write; `~/.config/cplt/` is the obvious place, and grants inside
+cplt's own state directory are refused for the same reason.
+
 ### Adding hosts with `allow.domains`
 
 The allowlist file is the right shape for a long, shared or subscription-fed
--- a/src/main.rs
+++ b/src/main.rs
@@ -2841,6 +2841,7 @@ fn start_proxy_if_enabled(
     config_path: Option<&PathBuf>,
     active_agent: agent::Agent,
     project_dir: &Path,
+    named_roots: &[PathBuf],
 ) -> anyhow::Result<Option<proxy::ProxyHandle>> {
     // Proxy-forced (#53): the proxy is mandatory. `with_proxy` defaults to true,
     // so the only way it is false here is an explicit disable (--no-proxy or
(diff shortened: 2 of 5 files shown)


Starefossen:
Four findings, three of them holes in the check itself. All fixed.

**The writable set was two entries short.** Named repositories are project-grade writable roots, and the system temp dirs and scratch dir are writable by construction with no grant to withdraw — a list file in any of them is as editable as one in the project. `sandbox::session_writable_roots` is now the single definition, sitting next to `writable_trees` so the two cannot drift on the trees no config creates.

**Canonicalizing the file first was itself a bypass** — the sharpest of the four. `<project>/list.txt -> /elsewhere/list.txt` resolved to a safe-looking target at launch, and the proxy re-reads through the *configured* path, so the agent swaps the symlink between reloads and owns the allowlist. The fix is to resolve the parent and leave the leaf alone: a symlinked spelling of the tree still matches, and a symlinked leaf is judged by where it sits rather than where it points today.

Worth saying that the assertion I wrote for this failed against my first attempt at the fix — the test found the weakness, not re-reading the code.

**The e2e fixture in `tempfile::tempdir()`** — correct, and it undercut the test: the system temp dir is writable by construction, so the fixture demonstrated the refused shape while claiming to show the supported one. It lives under `$HOME` now, which is deny-by-default and granted nothing here; the proxy reads the file in the unsandboxed parent, so reading still works.

**The displaced doc comment** — right, and it is the fifth time this pattern has bitten in this repo. `ScratchDir::base` moved below `create` so the numbered steps belong to the function they describe again.

One thing the review did not raise that its e2e finding exposed: the golden suite failed, because `config set` was accepting a list file the launch then refuses — the exact shape of #306. The same refusal now runs at write time with the same reason.

PR #518, opened 2026-09-15
fix: bound the gh auth token handover

Starefossen:
## Problem

`extract_gh_token()` (`src/sandbox_exec.rs`) runs `gh auth token --hostname github.com` through `Command::output()` with **no timeout**. It is reached from `configure_command` on every Copilot launch — the gh guard's `block_auth_token` defaults on — and it is the only unbounded wait cplt has between the startup banner and exec.

A `gh` that never answers therefore takes the whole launch down with it. The user sees

```
[cplt] Starting Copilot in sandbox...
```

and then nothing, with nothing on screen naming `gh` as the thing being waited on. Found while diagnosing a total hang at exactly that line.

## Fix

Bound the call at five seconds and degrade. No token means the agent's GitHub API calls fail, which is recoverable and visible; a hang means cplt never starts.

The wait reuses the kill-and-reap pattern already in the tree (`update::dpkg_owns`, `audit::git_output`): stdout drains on a helper thread so `gh` cannot block writing into a full pipe while the main thread polls, and on timeout the child is killed and reaped. The reader is detached rather than joined on that path — joining a thread that reads a pipe a surviving grandchild could still hold open would reinstate the hang this removes.

`stdin` is now closed explicitly. `Command::output()` gave the child a null stdin for free; a spawned child inherits the terminal instead, and a `gh` that decides to prompt would block on a read nobody answers.

`gh_auth_token(&Path, Duration)` is split out of `extract_gh_token()` so the timeout is testable without a five-second test.

## Tests

Three in `gh_token_extraction_tests`, all driving a `gh` stub:

- `a_gh_that_never_answers_does_not_block_the_launch` — `sleep 300`, 200 ms bound, must return `None` well inside 30 s.
- `a_gh_that_answers_still_hands_the_token_over` — the bound must not cost the feature it bounds.
- `a_failing_gh_hands_over_nothing` — non-zero exit with stdout, which `.trim()` alone would have accepted.

Proved red: with the old `output()` body restored, `a_gh_that_never_answers_does_not_block_the_launch` does not finish at all (killed at 25 s). With the fix the whole module runs in 0.32 s.

## Verification

```
cargo fmt -- --check                                                  clean
cargo clippy --all-targets -- -D warnings                             clean
cargo clippy --all-targets --target x86_64-unknown-linux-gnu -D warn  clean
cargo test --no-fail-fast (macOS, clone outside /tmp)                 0 failed
  lib 1178 · unit 356 · integration 188 (6 ignored) · e2e 152
  e2e_projects 64 · e2e_guards 72 · e2e_git_hardening 12
  network_audit 8 · others 89/13/4
```


Starefossen:
Both findings were right; fixed in 58a4014.

**Unbounded EOF wait.** The bound stopped one step short. `try_wait` reporting the child gone does not mean stdout has reached EOF — a descendant `gh` leaves behind holds the write end open. The reader now hands its buffer over a channel and is never joined on either path, and the post-exit wait gets a 500 ms grace of its own. Everything `gh` wrote is in the pipe by then, so the grace only bounds the descendant case. New test `a_descendant_holding_stdout_open_does_not_block_the_launch`: `gh` exits immediately with a backgrounded `sleep 300` on its stdout. Unbounded it blocks for as long as the descendant lives (10.3 s against a 10 s stub) and fails; bounded, the whole module runs in 0.81 s.

**Loose assertion.** Right — it proved the 30 s test guard worked and nothing about the bound. Now `waited < timeout + 2s` against the 200 ms timeout it passes, the margin covering process spawn and the 20 ms poll granularity on a loaded runner.

`audit::git_output` has the same latent shape and is deliberately left alone: it runs after the session rather than between the startup banner and exec, so it is a different risk and a different change.

PR #534, opened 2026-09-19
docs: explain settling with no-audit domain observation

RanaPriyansh:
`--no-audit --observe-domains` retains the descendant settle wait, but the flag interaction and warning were undocumented.

Document the two-second wait after direct-child exit, the warning when settling cannot be established, and the separate classification drain. Existing security caveats remain unchanged.

Verification: checked the wording against both execution paths, `AuditMode::for_run`, the settle probe, and existing descendant tests. `git diff --check` passed. No runtime tests or builds were run for this documentation-only change.

Refs https://github.com/navikt/cplt/pull/435#issuecomment-5668762340.

AI assistance was used to prepare and review this documentation change.


Starefossen:
Thanks, this is accurate against main. I pushed 69bf6dd on top with a few wording changes rather than asking for another round:

- Any audit suppressor takes this path, not only `--no-audit`: `--quiet` and `sandbox.audit = false` do too (`audit_enabled = resolved.audit && !resolved.quiet`).
- The probe only sees descendants that keep the inherited descriptor open (`src/audit.rs`, "a detector, not a cage"), so the paragraph now says a descendant that closes it is not detected. Otherwise silence reads as a guarantee.
- "Excluded" now names what from: the host list and `--observe-domains-out`.
- Moved the paragraph next to the `--observe-domains` text. The section it was in describes the default audit, which does not enable `--observe-domains`.

